PR #25, opened 2020-12-01
As a church member I can recover my password over email

MariamNaiga:
**What does this PR do?**

- Allows church members to recover/reset their password in the event that they have forgotten it.

**Description of tasks?**
- The user is able to reset their password through a unique link sent to them via email.
- The user's password should then be updated in the database.

**How can you test this manually?**
- Using Postman, and the link `localhost:4002/api/users/forgotPassword`, the user enters their email and sends a request:
`{
    "email": "*******@****.com"
}`
- The message returned will contain the user's unique token, and url of an email that will contain the reset link
- Send another request to `localhost:4002/api/users/resetPassword` with a body:
`{
  "token": " [ unique_token ] ",
  "password": " [new_password] "
}`
- The message returned will contain the url to the test confirmation email.

**Screenshot(s)**
Postman-forgotPassword
Postman-resetPassword
ResetPasswordEmail
ConfirmationEmail


tsconfig.json
```diff
@@ -9,7 +9,9 @@
     "sourceMap": true,
     "outDir": "./dist",
     "baseUrl": "./",
-    "incremental": true
+    "incremental": true,
+    //"esModuleInterop": true,
```
lytes20, line 13: Could you remove this commented out rule?

src/users/dto/reset-password.dto.ts
```diff
@@ -0,0 +1,10 @@
+import {IsNotEmpty} from 'class-validator';
+
+export class ResetPasswordDto {
+  @IsNotEmpty()
+  token: string;
+
+  @IsNotEmpty()
+  password: string;
+ 
+}
```
lytes20, line 10: Could you please add a new line at the end of this file

src/users/dto/update-user.dto.ts
```diff
@@ -5,8 +5,7 @@ export class UpdateUserDto {
     @IsNumber()
     id: number;
 
-    @IsNotEmpty()
-    roles: string[];
+    roles?: string[];
```
lytes20, line 8: Why did you remove the `@IsNotEmpty()` decorator?

MariamNaiga: For the scenario where the only thing I want to update is the password. With the `@IsNotEmpty()` on roles, the server will throw an error because the user hasn't put a value for roles.

PR #36, opened 2021-01-19
Edit Group Details

MariamNaiga:
**What does this PR do?**

- It adds the feature that allows team leaders to edit groups that they are leaders of.

**Description of tasks?**

- Team leaders can now edit groups that they are leaders of.

**How can you test this manually?**

- Send a `PUT` request to the address `localhost:4000/api/groups/group` with body:
```json
{
  "id": 0,
  "privacy": "Private",
  "name": "string",
  "details": "string",
  "categoryId": "string",
  "parentId": 0,
  "metaData": "string",
  "freeForm": "string",
  "latLon": "string",
  "placeId": "string"
}
```
- Response will be the details of the updated group

### DIFF
--- a/src/groups/services/groups.service.ts
+++ b/src/groups/services/groups.service.ts
@@ -5,7 +5,7 @@ import Group from '../entities/group.entity';
 import SearchDto from '../../shared/dto/search.dto';
 import { GroupSearchDto } from '../dto/group-search.dto';
 import { FindConditions } from 'typeorm/find-options/FindConditions';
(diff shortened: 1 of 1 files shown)


ekastimo (changes requested):
Is there a reason why you commented out, the place/location logic

MariamNaiga:
> Is there a reason why you commented out, the place/location logic

Yes, the reason I commented it out was because I get an internal server error when trying to update the group, and since the location of the group isn't being changed then there was no need to have that logic. Please correct me if I am wrong.

PR #37, opened 2021-01-29
Ref-011 Activate/Block User Feature

MariamNaiga:
**What does this PR do?**

- This PR adds the feature that allows administrators to activate or block users.

**Description of tasks?**

- Admins can now approve and block users. Blocked users will not be able to sign in with their credentials.

**How can I test this manually?**
- In Postman or Insomnia, make a `PUT request` to `localhost:4000/api/users` and give the request body: 
```json
{
  "id": 0,
  "roles": [
    "string"
  ],
  "password": "string",
  "isActive": true
}
``` 
- Database should update to make user active or inactive.

**ScreenShot(s)**

- Before Update
image

- PUT request
image

- After Update
image

### DIFF
--- a/src/auth/auth.helpers.ts
+++ b/src/auth/auth.helpers.ts
@@ -14,5 +14,6 @@ export const createUserDto = (user: User): UserDto => {
     fullName: getPersonFullName(user.contact.person),
     id: user.id,
     roles: user.roles,
+    isActive: user.isActive,
   };
 };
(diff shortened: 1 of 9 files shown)


androidcode33:
thanks @MariamNaiga is it possible to do a PATCH on the similar end-point? PUT works but have to send all the field else they are sent to null yet with PATCH you can only update the field(s) being updated.

PR #53, opened 2021-04-27
Ref-014, As MC admin, I can see which assigned members have attended their first MC meeting

dhotim:
**What does this PR do?**

- This PR adds a feature that allows MC admins to view first time guests. These are members who have attended an MC event for the first time.

**Description of tasks?**

- MC admins are now able to see new members who have attended an MC event.

**How to manually test this?**

- Under Reports, select a particular report to view then select members who have attended the MC. There should be a “First Time Guests” tab next to it to show members who have attended for the first time.

[Link to Client PR](https://github.com/kanzucode/angie-client/pull/50#issue-624690985)


src/events/event-attendance.controller.ts
```diff
@@ -125,6 +125,20 @@ export class EventsAttendanceController {
   async create(
     @Body() { id, ...data }: EventAttendanceCreateDto,
   ): Promise<EventAttendanceDto> {
+    const checkVisitor = await this.repository
```
ekastimo, line 128: This query gets all entries where this contact was a visitor, does not matter for which group.
I would expect `'"contactId"=:contactId AND "groupId"=:groupId', ` so that if there are 0 entries, then the person has never attended this group's events

ekastimo (changes requested):

src/events/event-attendance.controller.ts
dhotim: I have added that logic to the code. Members are only flagged as guests when it is their first time in a particular group.

image

ekastimo (approved):

PR #64, opened 2021-06-10
Add Sentry Application Monitoring

MariamNaiga:
**What does this PR do?**

- This PR adds an application monitoring feature, Sentry.

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -12,3 +12,4 @@ APP_URL=http://localhost:3000
 SENDGRID_API=your-api-key-here
 EMAIL_SENDER=example@email.com
 DATABASE_URL="postgresql://${DB_USERNAME}:${DB_PASSWORD}@${DB_HOST}:${DB_PORT}/${DB_DATABASE}?schema=public"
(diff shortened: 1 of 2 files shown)


ekastimo:
What exact logs are we really sending to sentry? Do you have a sample log sent to sentry for this

kakoma:
@MariamNaiga You'd missed adding `@sentry/node` and `@sentry/tracing` to `package.json`  Fixed and merged. Actually, please revise this to follow https://docs.sentry.io/platforms/javascript/guides/nextjs/ 

You just need to add `npm install --save @sentry/nextjs`, do the config and then  `import { withSentry } from '@sentry/nextjs';` and use it for our handlers. 

`export default withSentry(handler);`

Send a test as well...

```
// Next.js API route support: https://nextjs.org/docs/api-routes/introduction
import { withSentry } from '@sentry/nextjs';

const handler = async (req, res) => {
  throw new Error('API throw error test')
  res.status(200).json({ name: 'John Doe' })
}

export default withSentry(handler);
```

It should show up in the Sentry setup. Push the changes to this branch

#cc @ekastimo

PR #70, opened 2021-06-17
Fix-016. Multiple user accounts being created. Each user should have one account

dhotim:
**What does this PR do?**
- Adds a check to ensure that the email address entered is not already in use
- User is notified if the email address is already in use.

[Link to Client PR](https://github.com/kanzucode/angie-client/pull/70#issue-672314778)

### DIFF
--- a/src/crm/contollers/register.controller.ts
+++ b/src/crm/contollers/register.controller.ts
@@ -7,6 +7,7 @@ import { Repository } from 'typeorm';
 import { CreatePersonDto } from '../dto/create-person.dto';
 import { User } from '../../users/user.entity';
 import ContactListDto from '../dto/contact-list.dto';
+import { isEmptyObject } from 'src/utils/validation';
 
 @ApiTags('Register')
 @Controller('api/register')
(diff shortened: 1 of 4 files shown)


src/crm/contollers/register.controller.ts
```diff
@@ -21,6 +22,14 @@ export class RegisterController {
...
   async create(@Body() data: CreatePersonDto): Promise<ContactListDto> {
     const contact = await this.service.createPerson(data);
-    return ContactsService.toListDto(contact);
+    try {
+      return ContactsService.toListDto(contact);
+    } catch (e) {
+      if (isEmptyObject(contact)) {
+        return new ContactListDto();
+      } else {
+        return e;
```
ekastimo, line 31: Exceptions need to be thrown, not returned otherwise you will be getting 200 OK for a request that failed. I also added a comment on the client-side. there is a proper way to handle errors

ekastimo (changes requested):

src/crm/contollers/register.controller.ts
dhotim: > Exceptions need to be thrown, not returned otherwise you will be getting 200 OK for a request that failed. I also added a comment on the client-side. there is a proper way to handle errors

Thanks Timo for this. The code has been amended appropriately both here and client side.

ekastimo (approved):

Issue #82, opened 2021-06-21
As an admin, I'd like to fix the users/people list filters. Then for phone filter with a country code or not.


androidcode33:

androidcode33:
@dhotim add the PR URL here.

dhotim:
[Link to the the PR](https://github.com/kanzucodefoundation/project-zoe-server/pull/85#issue-674831617)

PR #85, opened 2021-06-21
Fix-002: As an admin, I'd like to fix the users/people list filters. Then for phone filter with a country code or not

dhotim:
# Ticket No
- #82 

# Summary of changes
- Split _churchLocations_ and _contactSearchDto_ to be defined with their own decorators
- Changed the client side logic for processing data into an array to be pushed to the endpoints via http.
- Specific changes made to ContactFilter.tsx and fitlerUtilities.ts at the client side

- Under the people menu, select more filters and search by any filter.
https://images.zenhubusercontent.com/234691966/23581e78-337d-4ea9-8e63-e91a7921685b/fix_002_people_filters.mp4

# Notes for reviewers
-  Validation logic has been restored at the dto. Changed the _initialData_ object key names to _cellGroup[]_ and _churchLocation[]_ at the client side. This was to enable a consistent way of pushing an array through to the endpoints successfully.

# Out of scope
-  Phone numbers need to be specifically formatted on insertion using international formats. An extra card should be created for this.

[Link to Client Side PR](https://github.com/kanzucodefoundation/project-zoe-client/pull/74#issue-677427691)

### DIFF
--- a/src/crm/dto/contact-search.dto.ts
+++ b/src/crm/dto/contact-search.dto.ts
@@ -10,13 +10,8 @@ export class ContactSearchDto extends SearchDto {
   @IsArray()
   @IsNumber({}, { each: true })
   @IsOptional()
-  cellGroups?: number[];
-
-  @Type(() => Number)
-  @IsArray()
-  @IsNumber({}, { each: true })
-  @IsOptional()
   ageGroups?: number[];
+  cellGroups?: number[];
   churchLocations?: number[];
   skipUsers?: boolean;
 }


src/crm/dto/contact-search.dto.ts
```diff
@@ -6,17 +6,12 @@ export class ContactSearchDto extends SearchDto {
   email?: string;
   phone?: string;
 
-  @Type(() => Number)
-  @IsArray()
-  @IsNumber({}, { each: true })
-  @IsOptional()
-  cellGroups?: number[];
-
```
ekastimo, line 14: Am a bit curious why you removed the validation logic

ekastimo (changes requested):

src/crm/dto/contact-search.dto.ts
dhotim: I have restored the validation logic and made some other changes at the processing of data at the client side.

PR #182, opened 2024-06-10
Feat: Search for Users, on the Manage Users page.

NAGERI:
# Ticket No

# Summary of changes
- Modified the results being returned, by `findAll ` in `user.service.ts` file.

- Search for a User using the email and name in the frontend. It should return results that meet the search terms.

# Notes for reviewers
- Check that the server runs without errors.

# Out of scope
-  none


NAGERI:
PR #183 is a correction of this.

PR #190, opened 2024-12-13
Add a matching algorithm to show duplicates in salvation records

masikotimo:
# Ticket No
- Team-Anchor-SalvationandFollowup-task-5
-  

# Summary of changes
- Added algorithm to match records and return them 

- 
Screenshot 2024-12-13 at 18 10 06

# Notes for reviewers
- Notes for reviewers here

# Out of scope
-  Out of scope here


kakoma:
Closing — the salvation entity has no tenantId and uses plain Repository instead of TenantAwareRepository, so it would break multi-tenancy. Salvation data is also now handled via the Report/ReportSubmission system. The concept is good but the implementation would need a full rewrite against current architecture.

PR #192, opened 2025-08-12
feat: Disabled SSL on local environment

NAGERI:
# Ticket No
- none

# Summary of changes
- Added a check, so if in the development environment, you can bypass the SSL connection while accessing your database.

- Check in the `.env` file that the `DB_ENABLE_SSL` value is set to false.
image

# Notes for reviewers
- You have to include `DB_ENABLE_SSL` in the  `.env` file.

# Out of scope
-  It is not related to metabase, please ignore the branch name.

### DIFF
--- a/src/shared/db.service.ts
+++ b/src/shared/db.service.ts
@@ -27,9 +27,13 @@ export class DbService {
         ...config.database,
         name: connectionName,
         type: "postgres",
-        ssl: {
-            rejectUnauthorized: false, // Required for DigitalOcean & Heroku
-        },
+        ssl:
+          process.env.DB_ENABLE_SSL === "false"
+            ? false
+            : {
+                rejectUnauthorized:
+                  false /* Required for DigitalOcean & Heroku */,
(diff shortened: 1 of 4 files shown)


src/shared/db.service.ts
```diff
@@ -27,9 +27,13 @@ export class DbService {
         ...config.database,
         name: connectionName,
         type: "postgres",
-        ssl: {
-            rejectUnauthorized: false, // Required for DigitalOcean & Heroku
-        },
+        ssl:
+          process.env.DB_ENABLE_SSL === "false"
```
kakoma: Switch this to this instead:
        ssl: process.env.DB_ENABLE_SSL === 'true' ? {
            rejectUnauthorized: false, // Required for DigitalOcean & Heroku
        } : false,
        
        The SSL logic should be added if the environment variable has been explicitly set as true

NAGERI:
@kakoma I have worked on it as suggested.

PR #195, opened 2025-11-15
Fix INSERT column mismatch in tenant migration

kakoma:
- Added tenantId column and Tenant relationship to GroupCategory entity
- Created tenants.helpers.ts to export Tenant entity for TypeORM
- Updated config.ts to include Tenant entity in appEntities array

This fixes the migration error "INSERT has more expressions than target columns" when migrating from schema-based to row-level tenancy.

# Ticket No
- Ticket number here

# Summary of changes
- Summary of changes here

# Notes for reviewers
- Notes for reviewers here

# Out of scope
-  Out of scope here


kakoma:
Explicit use of relation fields discouraged

PR #205, opened 2026-06-25
Fix/tasks and bulk upload of Contacts

NAGERI:
# Ticket No
- Matched ID 182 of the client

# Summary of changes
- fix(tasks): write ContactActivity when the task is marked unreachable

The UNREACHABLE status case was a no-op, so the retention report's
unreachable count always returned 0. Now records a ContactActivity
row so the report query has data to count.

Co-Authored-By: Claude Sonnet 4.6

# Notes for reviewers
- Some contacts, like children, do not have emails, but will be linked to the parents' contacts in future.

# Out of scope
-  Linking the child to the parent ( whose contact already exists in the system.

## Summary by CodeRabbit

* **New Features**
  * Added an **unreachable** contact activity type and included it in retention reporting outputs (overall, monthly, and weekly).
  * Retention reports now display **unreachable** metrics per period.
* **Bug Fixes**
  * Task status changes to **unreachable** now create the corresponding contact activity, improving tracking and reporting.
  * CSV contact imports no longer stop when email is missing; row matching now falls back to name + group.


kakoma:
@coderabbitai full review

kakoma (approved):

PR #211, opened 2026-06-28
fix(groups): prevent active leaders from being added as members

kakoma:
A leader GroupMembership role could be silently overwritten to Member when re-adding an inactive leader via the batch membership endpoint. Added a guard in GroupsMembershipService.create that rejects the request with a 400 if any contact in the batch is already an active leader of the group. Also fixes stale mock setup and outdated assertions in the spec.

# Ticket No
- Ticket number here

# Summary of changes
- Summary of changes here

# Notes for reviewers
- Notes for reviewers here

# Out of scope
-  Out of scope here

## Summary by CodeRabbit

* **Bug Fixes**
  * Added validation to prevent adding a member when that contact is already an active leader in the same group.
  * Re-adding a previously inactive leader now works as expected.
  * Updated membership results and logging to reflect upsert behavior, including created/reactivated counts and affected contacts.
* **Tests**
  * Expanded service coverage for new membership rules and updated group lookup handling.

### DIFF
--- a/src/groups/services/group-membership.service.spec.ts
+++ b/src/groups/services/group-membership.service.spec.ts
@@ -5,19 +5,19 @@ import GroupMembership from '../entities/groupMembership.entity';
 import Group from '../entities/group.entity';
 import { AppLogger } from '../../utils/app-logger.service';
 import { GroupRole } from '../enums/groupRole';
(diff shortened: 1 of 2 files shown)


kakoma:
@coderabbitai full review

PR #217, opened 2026-06-30
Fix/report submission resolve contact names

NAGERI:
Fields backed by the 'dynamic_member_selector' widget store a Postgres
    array of contact IDs as their fieldValue (e.g. '{"51","58","26"}').
    Replace that raw value with a comma-separated list of contact names.
   

# Ticket No
- Ticket number here

# Summary of changes
- Summary of changes here

# Notes for reviewers
- Notes for reviewers here

# Out of scope
-  Out of scope here

### DIFF
--- a/src/groups/services/group-membership.service.ts
+++ b/src/groups/services/group-membership.service.ts
@@ -70,7 +70,23 @@ export class GroupsMembershipService {
       take: req.limit ?? 100,
       where: filter,
     });
-    return data.map((it) => this.toDto(it, req.groupId));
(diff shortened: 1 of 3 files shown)


kakoma:
@NAGERI "Replace that raw value with a comma-separated list of contact names." Don't mix up what should be a presentation layer issue with a storage issue. The items should still be stored as IDs but the presentation layer should take those IDs and present them as names; essentially, when retrieving the data for display, replace the IDs with names.

PR #222, opened 2026-07-02
fix(membership): bulk group membership creation success message response

iamMarybrightEE:
# Ticket No
- **TICKET-1237**

# Summary of changes
- Modified the success JSON response payload structure inside `GroupMembershipController` for batch user creation.
- Customized the return confirmation string to read explicitly as `"Members added successfully"` instead of generic operational status updates.

# How should this be tested? 
- Login, navigate to the group details page and try to add members.
- Confirm that the server returns a successful `201 Created` status code and that the response message is "Members added successfully"

# Notes for reviewers
- This code patch changes only the text returned to the client application interface inside the route handler level wrapper. It has completely zero impact on relational database query integrity or entity constraints.

### DIFF
--- a/src/groups/controllers/group-membership.controller.ts
+++ b/src/groups/controllers/group-membership.controller.ts
@@ -38,7 +38,7 @@ export class GroupMembershipController {
   async create(@Body() data: BatchGroupMembershipDto): Promise<any> {
     const created = await this.service.create(data);
     return {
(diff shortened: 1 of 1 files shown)


iamMarybrightEE:
@NAGERI

kakoma (approved):

PR #223, opened 2026-07-06
fix(crm): allow contact search to match full names with spaces

iamMarybrightEE:
# Ticket No
- CRM-382 

# Summary of changes
- Replaced column-specific `ILike` array filters with a dynamic `QueryBuilder` implementation inside `ContactsService.findAll`.
- Tokenized the incoming search string (`req.query`) by word boundaries to match multi-word phrases against `firstName`, `lastName`, and `middleName` simultaneously.
- Switched data extraction to `getRawMany()` to provide a flat, predictable results layout for safe array intersection operations.

# How should this be tested?
1. Open the people  page.
2. Search for a single term (e.g., `"Sarah"`) and confirm matching results show up.
3. Search for a combined full name containing spaces (e.g., `"Sarah Nambi"`) and press Enter.
4. Verify that the correct contact maps and returns successfully instead of throwing an empty state.

# Out of scope
- Refactoring the telephone (`req.phone`) or email (`req.email`) string processing code blocks.
- Modifying frontend input keypress debouncer delays.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved contact name search to handle multi-word queries more accurately across first, middle, and last names.
  * Search now evaluates each term separately, improving precision for full-name and partial-name matches.
  * Enhanced handling of special characters in search input, reducing cases where searches fail or return incorrect matches.

### DIFF
--- a/src/crm/contacts.service.ts
+++ b/src/crm/contacts.service.ts
@@ -94,6 +94,10 @@ export class ContactsService {
     this.logger = this.appLogger.createContextLogger('ContactsService');
   }
 
+  private static escapeLike(value: string): string {
+    return value.replace(/[%_\\]/g, (match) => `\\${match}`);
+  }
+
   async findAll(req: ContactSearchDto, user?: any): Promise<ContactListDto[]> {
     const tracking = this.logger.startTracking('findAllContacts', {
       userId: user?.id,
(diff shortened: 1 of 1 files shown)


iamMarybrightEE:
@NAGERI review completed

kakoma (approved):

PR #226, opened 2026-07-08
feat(users): support fellowship leaders without email addresses

kakoma:
Add a nullable email column to User, decoupling the login identifier (username) from the email address. Email-less users are created with an admin-supplied username (e.g. phone number); password reset is gracefully declined for accounts with no email on file.

- User entity: add email: string | null (nullable, unique column)
- CreateUserDto: make username optional (used when contact has no email)
- UserListDto / auth UserDto: surface email as nullable
- createUser: derive username+email from contact email record when present; fall back to data.username for email-less contacts; skip welcome email if null
- register: set user.email alongside user.username
- createUserDto helper: map email from user.email, not user.username
- forgotPassword: guard on null email, return clear admin-contact message
- resetPassword: use user.email as To: address; skip send if null

# Ticket No
- Ticket number here

# Summary of changes
- Summary of changes here

# Notes for reviewers
- Notes for reviewers here

# Out of scope
-  Out of scope here

## Summary by CodeRabbit

* **New Features**
  * User profiles and lists now include an email address when available.
  * Account creation and password reset emails are now sent to the saved email address.

* **Bug Fixes**
  * Email-related flows now handle missing email addresses more gracefully.
  * Password reset and welcome messages no longer rely on a username as the email recipient.

* **Documentation**
  * API responses now reflect that email can be optional or unavailable.

### DIFF
--- a/src/auth/auth.helpers.ts
+++ b/src/auth/auth.helpers.ts
@@ -11,7 +11,7 @@ export const createUserDto = (user: User): UserDto => {
   user.userRoles.forEach((it) => permissions.concat(it.roles.permissions));
   return {
     contactId: user.contact.id,
-    email: user.username,
+    email: user.email ?? null,
     username: user.username,
     fullName: getPersonFullName(user.contact.person),
     id: user.id,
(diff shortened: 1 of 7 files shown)


NAGERI:
@coderabbitai full review

PR #236, opened 2026-07-10
Fix/tasks location group ids qs array limit

NAGERI:
Root cause: `qs (Express/Nest's query-string parser)` only decodes repeated query keys like `locationGroupIds=1&locationGroupIds=2...` into an array up to its default arrayLimit of 20 entries. Past that, it silently falls back to a plain object `({"0": "142", "1": "70", ...}).` The controller's `Array.isArray(locationGroupIds)` check then failed, wrapped the whole object into a single-element array, and Number(object) produced NaN — the exact single NaN parameter seen crashing the SQL query in `tasks.service.ts:178.`

Fix ([tasks.controller.ts:46-80](vscode-webview://0r2qf483ilgpe7kp3ckdl10b3q080e006j1b2m8rkfqo4poeqdbf/src/tasks/tasks.controller.ts#L46-L80)): widened the param type to include the object shape and added a toNumberArray helper that handles all three shapes (string, string[], or the qs fallback object), converting via Object.values() when needed and filtering out any non-finite results before they ever reach tasks.service.ts.

# Ticket No
- Ticket number here

# Summary of changes
- Summary of changes here

# Notes for reviewers
- Notes for reviewers here

# Out of scope
-  Out of scope here

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved contact import group resolution by applying the current user’s access context.
  * Fixed task filtering when location group IDs are provided in different query formats.
  * Invalid or empty location group values are now ignored instead of producing unintended filters.

### DIFF
--- a/src/crm/contollers/contact-import.controller.ts
+++ b/src/crm/contollers/contact-import.controller.ts
@@ -123,6 +123,7 @@ export class ContactImportController {
           const groupData = await this.groupsService.findOne(
             effectiveGroupId,
             false,
+            req.user,
           );
           if (!groupData) {
             throw new BadRequestException({
--- a/src/crm/contollers/contact-import.controller.ts
+++ b/src/crm/contollers/contact-import.controller.ts
@@ -173,7 +173,10 @@ export class ContactImportController {
 
   @Post('groupLeaders')
   @UseInterceptors(FileInterceptor('file'))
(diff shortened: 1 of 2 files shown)


kakoma:
Superseded by #237 and #239, which include the same qs array-limit fix with the stricter `Number.isInteger(id) && id > 0` filter plus additional fixes. All changes from this PR are already in develop.

PR #243, opened 2026-07-15
Fix: MC leader dropdown showing all groups instead of only their subtree

iamMarybrightEE:
# Fix: MC leader dropdown showing all groups instead of only their subtree

## Problem
Leaders saw **every MC group** in the report-submission dropdown, instead of just:
- Their own group, if they lead a leaf MC
- Their group + all child MCs, if they lead a parent MC

## Root Cause
`getUserIsMemberLeaderGroupIds()` (in `GroupPermissionsService`) queried memberships with:

```typescript
role: In([GroupRole.Member, GroupRole.Leader]),
```

A user who was merely a **Member** of a group (not a Leader) still had that group's ID expanded into its full descendant subtree. If that group sat high up the tree, the user inherited visibility into every descendant MC underneath — despite having no leadership role there.

Since only **Leaders** should have report-submission access, Member rows should never trigger subtree expansion.

## Fix
- Filtered membership lookups to `GroupRole.Leader` only.
- Removed duplicate logic: `getUserIsMemberLeaderGroupIds` now delegates to `getUserGroupIds` (already leader-only, already does descendant expansion) instead of re-implementing the same query, expansion, and deduplication.
- Renamed the method to `getUserIsLeaderGroupIds` to drop the now-inaccurate "Member" wording, and updated its call site in `getMyGroups()`.
- Wrapped the `Group` repository in `TenantAwareRepository` so group lookups in this service are properly tenant-scoped (previously built from a raw `connection.getRepository`).

## Files changed
- `group-permissions.service.ts`
- `group-permissions.service.spec.ts`
- `groups.service.ts`

## Known limitations / follow-ups (not addressed in this PR)
- **`GroupMembership` is not tenant-scoped.** The entity has no `tenant` column or relation, so `membershipRepository` remains a plain repository. Membership-based permission checks (`hasPermissionForGroup`, `getUserGroupIds`) rely on `contactId` being globally unique for tenant isolation. If contacts can collide across tenants, this needs a schema change to add direct tenant scoping to `GroupMembership` — out of scope here.
- Confirmed no other call sites depend on the old Member-inclusive behavior of the renamed method. If one turns up later, it will need explicit handling rather than silently inheriting leader-only scoping.

## Testing
- [x] Leaf MC leader sees only their own group in the dropdown
- [x] Parent MC leader sees their group + all child MCs
- [x] Plain Members (non-leaders) no longer see any groups via this endpoint
- [x] Existing leader flows (report submission, group member views) unaffected
- [x] Unit tests updated: leader-only membership behavior, descendant expansion via delegation, tenant-aware repository construction args

## Summary by CodeRabbit

- **Bug Fixes**
  - Group permission checks now respect the active tenant, helping prevent cross-tenant access.
  - Leader-based group access now consistently includes descendant groups.
  - Access results for leader-specific checks are limited to groups where the user has a leader role.
- **Tests**
  - Added coverage for tenant-aware permission handling and leader descendant access.

### DIFF
--- a/src/groups/services/group-permissions.service.spec.ts
+++ b/src/groups/services/group-permissions.service.spec.ts
@@ -6,12 +6,17 @@ import GroupMembership from '../entities/groupMembership.entity';
 import { GroupRole } from '../enums/groupRole';
 import ClientFriendlyException from '../../shared/exceptions/client-friendly.exception';
 import { roleAdmin } from '../../auth/constants';
+import { TenantContext } from '../../shared/tenant/tenant-context';
+import { TenantAwareRepository } from '../../shared/repository/tenant-aware.repository';
+
+jest.mock('../../shared/repository/tenant-aware.repository');
 
 describe('GroupPermissionsService', () => {
   let service: GroupPermissionsService;
   let mockMembershipRepo: any;
   let mockGroupRepo: any;
-  let mockTreeRepo: any;
+  let mockConnection: any;
+  let mockTenantContext: any;
 
   const adminUser = { contactId: 1, roles: [roleAdmin.role] };
   const regularUser = { contactId: 2, roles: [] };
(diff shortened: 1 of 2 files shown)


iamMarybrightEE:
@NAGERI

NAGERI (approved):

PR #246, opened 2026-07-16
fix: replace closure-table lookup with parentId walk in getGroupAndAllChildren

iamMarybrightEE:
## What
Fixes broken report visibility for Location Pastors and Regional Leaders, who could no longer see MC reports within their location/region.

## Root cause
`GroupTreeService.getGroupAndAllChildren` relied on TypeORM's closure-table tree strategy (`treeRepository.findDescendants`) to expand a user's group into all its descendants. The closure table (`group_closure`) is a separate structure that TypeORM only keeps in sync when groups are reparented through the ORM's `parent`/`children`
relations. If a group's `parentId` was ever updated another way (raw SQL, a bulk script, a migration), the closure table silently drifts out of sync for that branch — while `parentId` itself stays correct.

This meant report visibility (which depends on `getGroupAndAllChildren`) could silently lose descendant groups, while other permission checks elsewhere in the codebase — which already use a manual `parentId` walk in `GroupPermissionsService.getGroupAndAllDescendants` — kept working correctly.

## Change
Rewrote `getGroupAndAllChildren` to walk `parentId` manually via BFS, instead of reading from the closure table. Same function signature, same caching behavior, same callers — only the internal traversal logic changed.

```ts
// before
const descendants = await this.treeRepository.findDescendants(group);

// after
const children = await this.repository.find({
  where: { parentId: currentGroupId },
  select: ['id'],
});
```

This removes the dependency on a data structure that's proven to drift, and brings this code path in line with the traversal logic already working correctly elsewhere in the codebase.

## Testing
- Verified in staging: Location Pastor now sees all MC reports within   their location.
- Verified in staging: Regional Leader now sees all reports within   their region.
- Confirmed cache keys/behavior unchanged (still 30-min TTL, same   key format).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when loading groups and their full descendant hierarchy.
  * Reduced database work during hierarchical expansion for faster, more consistent performance.
  * Preserved existing caching behavior to keep repeated loads quick.
* **Tests**
  * Refactored and expanded unit tests to validate child/descendant retrieval, caching, and graceful handling of missing groups.
  * Added assertions ensuring the descendant lookup path is exercised without relying on tree-repository descendant queries.

### DIFF
--- a/src/groups/services/group-tree.service.spec.ts
+++ b/src/groups/services/group-tree.service.spec.ts
@@ -89,7 +89,8 @@ describe('GroupTreeService', () => {
     });
     it('does not rely on treeRepository for descendant lookup', async () => {
       mockGroupRepo.find.mockImplementation(({ where }: any) => {
-        if (where.parentId === 10) return Promise.resolve([{ id: 11 }]);
+        const ids = extractIds(where.parentId);
+        if (ids.includes(10)) return Promise.resolve([{ id: 11 }]);
         return Promise.resolve([]);
       });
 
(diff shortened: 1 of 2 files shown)


iamMarybrightEE:
@coderabbit full review

iamMarybrightEE:
@coderabbit full review

iamMarybrightEE:
@coderabbit full review

iamMarybrightEE:
@coderabbit full review

iamMarybrightEE:
@NAGERI review complete

NAGERI (approved):

PR #255, opened 2026-07-30
feat(users,groups): scope /api/users to a location group, include FOB leaders

iamMarybrightEE:
## Summary of changes
- `UsersService.findUsersInGroup(groupId)` — resolves the group's   active members, walks up to find the nearest ancestor group whose   category name is `FOB` (if any), and returns the union of that  group's members plus the FOB's active Leaders, in one membership  query using an OR condition (not two separate queries/endpoints).
- `UsersController` — new `GET /api/users/by-location/:groupId`.
- `GroupsService.getContactLocationGroup(contactId)` — thin wrapper  around the existing `GroupPermissionsService.getContactLocationGroupId`,  returning `{ id, name }` or `null`.
- `GroupController` — new `GET /api/groups/contact-location/:contactId`.
- `TasksService.findTaskWithRelations` — now calls the existing  `attachLocationGroups` helper (previously only called from  `findAll`/`findAllForContact`), so single-task fetches also carry  `locationGroup`. This was the root cause of the `/by-location/undefined`  400s seen during testing.

## How should this be tested
1. Seed/confirm a hierarchy: FOB → ... → Location, with a   `GroupMembership` of role `Leader` on the FOB group for some user's   contact, and separate members directly on the Location group.
2. `GET /api/groups/contact-location/:contactId` for a contact in that   location → expect `{ id, name }` of the location group.
3. `GET /api/users/by-location/:locationGroupId` → expect both the   direct location members and the FOB's Leader(s), no duplicates,   in one response.
4. Contact/location with no FOB ancestor → same endpoint returns only   direct location members, unchanged from pre-FOB behavior.
5. `GET /api/tasks/:id` (single task fetch), and the task update /   reassign / status-update / create flows → confirm each response   now includes `task.locationGroup` populated, not just the list   endpoints.
6. Regression: confirm `findAll` and `findAllForContact` responses are   unaffected (they already called `attachLocationGroups`, untouched   here).

## Note for reviewers
- The FOB match is a literal string check on `group.category.name ===  'FOB'`. Please confirm this matches your actual seed/category data —   if the category is named differently, this needs a one-line update.
- The ancestor walk is a sequential loop of `findOne` calls (bounded by  tree depth, with cycle protection via a `visited` set). Fine for  current hierarchy depths; flagged as a candidate for a single  recursive query if this becomes a hot path.
- `findTaskWithRelations` change affects `findOne`, `create`, `update`,  `updateStatus`, and `reassign` simultaneously since they all share  this helper — worth a broader regression pass on those flows, not  just the assignment-related ones.

## Out of scope
- No new authorization checks — the location/FOB-scoped user list is  returned regardless of whether the requesting user has access to  that location/FOB. If that's a concern, it's a separate follow-up.
- No changes to `GroupPermissionsService.getContactLocationGroupId`  itself, reused as-is.
- No migration or seed-data changes for FOB categories or leader  memberships.

## Summary by CodeRabbit

* **New Features**
  * Added location-group lookup for contacts.
  * Added an endpoint to list users in a location group, including relevant leaders.
  * Enhanced task details with location-group information and contact address data.
* **Improvements**
  * Improved location-based access checks for tasks and group information.
  * Added stronger validation for group members, permissions, pagination, and SMS operations.
  * Improved handling of missing or restricted location groups.
* **Tests**
  * Expanded coverage for groups, users, tasks, permissions, pagination, and SMS scenarios.

### DIFF
--- a/src/groups/controllers/group.controller.spec.ts
+++ b/src/groups/controllers/group.controller.spec.ts
@@ -4,6 +4,7 @@ import { GroupPrivacy } from '../enums/groupPrivacy';
 import { GroupsService } from '../services/groups.service';
 import { GroupController } from './group.controller';
 import { UsersService } from '../../users/users.service';
+import ClientFriendlyException from '../../shared/exceptions/client-friendly.exception';
 
 describe('GroupController', () => {
   let controller: GroupController;
@@ -78,7 +79,10 @@ describe('GroupController', () => {
       };
     }),
     getContactLocationGroup: jest.fn(
-      async (contactId: number): Promise<{ id: number; name: string } | null> => {
+      async (
+        contactId: number,
+        user: any,
+      ): Promise<{ id: number; name: string } | null> => {
         return { id: 7, name: 'Downtown' };
       },
     ),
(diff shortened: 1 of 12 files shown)


iamMarybrightEE:
@coderabbit nitpick 2: I'm keeping the per-level parentId walk instead, because that's exactly the call that caused the "public" alias was not found production error.

iamMarybrightEE:
@NAGERI review complete

NAGERI (approved):

PR #257, opened 2026-08-06
fix dashboard birthday locations

iamMarybrightEE:
# Summary of changes
- Fixed `DashboardService.getBirthdaysThisWeek()` incorrectly labeling a contact's **location** with their **Missional Community (MC)** name instead of their actual campus/venue.
  - Root cause: the method read `activeMembership.group.name` directly off a contact's `GroupMembership`, but memberships are recorded at the fellowship (MC) level, not the location level. `Group` has no `type`/`level` column of its own — hierarchy level is expressed via `Group.category → GroupCategory.purpose`.
- Added `loadGroupHierarchyMap()` to load the tenant's full group tree (`id`, `parentId`, `category.purpose`, `name`) once per request, avoiding N+1 queries.
- Added `resolveLocationName()` / `resolveLocationNames()` to walk `parentId` upward from a contact's active membership(s) to the nearest ancestor whose `category.purpose === GroupCategoryPurpose.LOCATION`.
- **Multiple active memberships are now all considered**, not just the first found — a contact belonging to locations A and B now shows `"A, B"` instead of an arbitrary single pick (or a silent fallback if the first-found chain happened to be unresolvable).
- Replaced the `'General Locations'` placeholder fallback with `resolveTenantRootName()`, which returns the tenant's actual root **group** name (the group with no `parentId`) — used only when a contact has no active memberships or none resolve to a `LOCATION`-purpose ancestor. Falls back to `'Unknown'` only if the tenant's group tree is empty/unreachable, as a last-resort guard.
- Explicitly ruled out using `Tenant.name`/a new `Tenant.displayName` for this fallback — `Tenant.name` is a lower-cased, space-stripped slug (`worshipharvest`) with no reliable way to recover proper casing/spacing algorithmically, and adding a migration was out of scope for this fix.
- Fixed a date-parsing bug: bare `'YYYY-MM-DD'` values from `Person.dateOfBirth` were being parsed via `new Date(string)`, which JS treats as **UTC midnight** — this can shift the resulting local date backward by one day depending on server timezone. Added `parseDateOnly()` to parse date-only strings as local-midnight instead.
- Cleaned up type/lint issues:
  - Removed inline `any` return type from `getBirthdaysThisWeek()` in favor of an explicit `{ birthdays: BirthdayEntry[] }` return type.
  - Extracted the repeated hierarchy-entry shape into a single `GroupHierarchyEntry` interface instead of five inline duplicates.
  - Removed a silent `catch (error) { return null }` (unused `error` binding, no observability) in favor of `this.logger.warn(...)` with the contact ID and error message, while preserving the same skip-and-continue behavior.
  - Merged the separate `.filter(contact => contact.person?.dateOfBirth)` + `.map()` into a single `.map()` with an early-return guard, so `person` is correctly narrowed to non-optional within the closure that builds the birthday object (previously a latent strict-mode type gap).

# How should this be tested?
- Hit the birthdays endpoint (`getBirthdaysThisWeek`) for a tenant with:
  1. A contact whose only active membership is an MC → `location` should show the MC's parent **Location** name, not the MC name.
  2. A contact with two active memberships under two different Locations → `location` should show both, comma-separated.
  3. A contact with no active memberships (or whose chain never reaches a `LOCATION`-purpose group) → `location` should show the tenant's root group name, not `'General Locations'`.
  4. A contact whose `dateOfBirth` is stored as a bare `'YYYY-MM-DD'` string → confirm the returned `upcomingDate` lands on the correct calendar day for a server running in a non-UTC timezone (e.g. test with server TZ set to something like `America/New_York` and a birthday near midnight UTC boundary).
- Confirm no N+1 queries: `loadGroupHierarchyMap` should fire once per request regardless of contact count (check query logs).
- Confirm malformed/unparseable `dateOfBirth` values are still skipped silently from the response but now emit a `logger.warn` line with the contact ID.

# Notes for reviewers
- No migration included — this deliberately avoids adding `Tenant.displayName` (discussed and dropped in favor of using the existing group hierarchy, which required no schema changes).
- `resolveTenantRootName` currently joins multiple roots with `, ` if a tenant's groups form more than one disconnected tree (no single group with `parentId: null`... multiple). This is an edge case that shouldn't happen in normal data but is handled defensively rather than assumed away.
- Happy to split the date-parsing fix (`parseDateOnly`) into its own PR if you'd rather keep this change scoped strictly to the location-resolution bug — flagging it here since it was found in the same block during review.

# Out of scope
- Persisting a human-friendly tenant/church display name (e.g. `"Worship Harvest"` vs. the stored slug `worshipharvest`) — would require a migration and a decision on where that value is captured (signup flow vs. admin-editable field). Discussed but deferred.
- Backfilling or fixing existing `GroupCategory` rows with a `null` `purpose` — contacts under those categories will fall back to the tenant root name until those categories are stamped with the correct purpose.
- Any changes to `GroupPermissionsService` or dashboard-level (Sunday Service) location scoping — this PR only touches `getBirthdaysThisWeek`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved birthday listings by considering only active memberships.
  * Added more accurate location and hierarchy information, with tenant-level fallback names when needed.
  * Improved handling and validation of date-only birthdays.
  * Ensured malformed birthday records are handled without disrupting results.
  * Standardized birthday date formatting for consistent display.

### DIFF
--- a/src/dashboard/dashboard.service.ts
+++ b/src/dashboard/dashboard.service.ts
@@ -1,4 +1,4 @@
-import { Injectable, Inject, NotFoundException } from '@nestjs/common';
+import { Injectable, Inject, NotFoundException, Logger } from '@nestjs/common';
 import { Between, Connection, In, Repository } from 'typeorm';
 import { ReportSubmission } from '../reports/entities/report.submission.entity';
 import { Report } from '../reports/entities/report.entity';
@@ -8,6 +8,7 @@ import Person from '../crm/entities/person.entity';
 import Contact from '../crm/entities/contact.entity';
 import { TenantContext } from '../shared/tenant/tenant-context';
 import { ContactCategory } from '../crm/enums/contactCategory';
+import { GroupCategoryPurpose } from '../groups/enums/groups';
 
 interface PgaTrendPoint {
   period: string;
@@ -20,6 +21,22 @@ interface LocationPgaRanking {
   pga: number;
 }
 
+interface GroupHierarchyEntry {
+  parentId: number | null;
+  purpose: GroupCategoryPurpose | null;
+  name: string;
+}
+
+interface BirthdayEntry {
+  id: number;
+  name: string;
+  dateOfBirth: string | Date;
+  location: string;
+  upcomingDate: string;
+}
+
+const UNKNOWN_ROOT_LABEL = 'Unknown';
+
 @Injectable()
 export class DashboardService {
   private readonly reportSubmissionRepository: Repository<ReportSubmission>;
@@ -27,6 +44,7 @@ export class DashboardService {
   private readonly groupRepository: Repository<Group>;
   private readonly personRepository: Repository<Person>;
   private readonly contactRepository: Repository<Contact>;
+  private readonly logger = new Logger(DashboardService.name);
 
   constructor(
     @Inject('CONNECTION') connection: Connection,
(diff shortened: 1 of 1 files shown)


iamMarybrightEE:
@NAGERI review complete

NAGERI (approved):

PR #266, opened 2026-08-07
Feat/submissions compliance

iamMarybrightEE:
# Summary of changes
- Adds `ReportsService.getMcSubmissionCompliance(user, from?, to?)`, which computes MC Attendance Report submission status per MC group/leader across a range of reporting weeks (Sunday-start).
- Scopes results to the caller's manageable groups via the existing `getUserManageableGroups(user)` helper (same scoping used by `getWeeklyMcaSummary` and `getMyGroupsSubmissions`) so leaders only ever see MCs within their own hierarchy.
- Only returns groups with at least one missed reporting week in range — fully compliant groups are excluded from the response entirely, since this endpoint answers "who hasn't submitted," not "everyone's status."
- Filters the MC group lookup by the group category's `GroupCategoryPurpose` (fellowship) rather than the mutable `GroupCategoryNames.MC` display name, so renaming a category no longer silently breaks compliance scoping.
- Adds `GET /reports/mc/compliance?from=&to=` route, following the same controller conventions as the existing `GET /reports/mca/weekly-summary` route (`@Header('Cache-Control', 'no-store')`, `apiLog`, `request.user`).
  - `from`/`to` query params are parsed as local-calendar `YYYY-MM-DD` dates (not `new Date(string)`, which treats date-only strings as UTC midnight and shifts the day for any timezone west of UTC). Malformed dates and a `from` after `to` are rejected with a `400`.
- Refactors `fetchMemberSelectorContactNames`'s inline contact-name query into a shared `fetchContactNames(contactIds)` helper, reused by the new `getMcLeaderNamesByGroupId` to avoid duplicating the contact/person query. This query is tenant-scoped (`c.tenantId = :tenantId`).
- Adds `COMPLIANCE_MAX_WEEKS_LOOKBACK = 12` as the default lookback window when `from` is omitted (i.e. the "all time" filter), to keep the query bounded and the result meaningful.
- **Timezone/date-boundary fix (bundled into this PR because `getMcSubmissionCompliance` shares the same week-boundary logic as `getWeeklyMcaSummary`, and both need to be correct together):**
  - `getStartOfWeek` previously derived the week-start `Date` using local-time getters (`setHours`, `getDay()`), then callers formatted it with `.toISOString().slice(0, 10)`. For any tenant running ahead of UTC (e.g. Africa/Kampala, UTC+3), converting a local midnight to UTC silently rolls the date back a day, so the `reportingPeriod` string written at submit time did not reliably match the calendar week a submission actually belonged to.
  - Replaced the ad hoc `.toISOString().slice(0, 10)` formatting throughout the file with a single `formatDateKey(date)` helper that reads **local** calendar fields (`getFullYear`/`getMonth`/`getDate`), so week-boundary strings are computed consistently everywhere they're used (`submitReport`, `getWeeklyMcaSummary`, `getMcSubmissionCompliance`).
  - More importantly: **`getWeeklyMcaSummary` and `getMcSubmissionCompliance` no longer match against the stored `reportingPeriod` string at all.** They now query the real `submittedAt` timestamp column directly (`Between`/range comparison) and bucket submissions into weeks live, at read time, using the current `getStartOfWeek`. This makes both endpoints immune to any future change in how `reportingPeriod` is computed or formatted — a submission's week is always derived from its actual timestamp, not from a precomputed string that could go stale if the formula ever changes again. `reportingPeriod` itself is left untouched for `submitReport`'s duplicate-submission-per-week check, which is a write-time concern and unaffected by this.
  - Added regression tests: `getWeeklyMcaSummary` now includes a submittedAt-boundary test (submission at 23:59:59 on the last day of the week must still count, guarding the exclusive-upper-bound query), and `getMcSubmissionCompliance` includes a test asserting a submission with a deliberately stale/wrong `reportingPeriod` is still bucketed correctly via its `submittedAt`.
- Adds unit tests covering: visibility scoping (no manageable groups → empty result, never queries MC groups), missing MC Attendance Report handling, exclusion of fully-compliant groups, correct sorting/aggregation of partially-compliant groups, `missingCurrentWeek` flag correctness, the "No leader assigned" fallback, controller-level date parsing/validation, and the two date-boundary regressions above.

1. Run the new test suite: `describe('getMcSubmissionCompliance', ...)` and the updated `describe('getWeeklyMcaSummary', ...)` in `reports.service.spec.ts`.
2. Manually hit `GET /reports/mc/compliance` as a user who manages a subset of MC groups and confirm the response only contains those groups, never groups outside their hierarchy.
3. Manually hit the same endpoint as a user with no manageable groups and confirm `{ weekStarts: [...], groups: [] }` is returned, not an error.
4. Hit `GET /reports/mc/compliance?from=2026-13-40` (and other malformed/out-of-order date combinations) and confirm a `400` is returned rather than a silently wrong date being used.
5. Submit an MC Attendance Report for a given group/week, then confirm that group drops out of the compliance response for that week on the next call, **and** that `GET /reports/mca/weekly-summary` reflects it too — test this on a server where `TZ` is not `UTC` (e.g. `Africa/Kampala`) to exercise the actual bug this PR fixes.
6. Test `from`/`to` omitted entirely and confirm the lookback is capped at 12 weeks rather than scanning all history.
7. Confirm a group with no active leader membership returns `leaderName: 'No leader assigned'` rather than throwing.
8. Submit a report late in the week (e.g. Saturday night, local time) and confirm it's still counted in that week's summary/compliance rather than silently attributed to the wrong week.
9. Rename an MC category's display name in a test tenant and confirm `getMcSubmissionCompliance` scoping is unaffected (still resolves by `GroupCategoryPurpose`, not the name string).

# Notes for reviewers
- `getMcSubmissionCompliance` deliberately filters out compliant groups before returning (`.filter((group) => group.weeksMissed > 0)`) — this was a fix from an earlier draft that returned everyone's status and let the frontend decide, which was the wrong place to encode "not submitted" semantics.
- Visibility relies entirely on `getUserManageableGroups(user)` already being correct/trusted — this PR doesn't change that helper, only reuses it, consistent with `getWeeklyMcaSummary`.
- `getMcLeaderNamesByGroupId` assumes at most one meaningful "leader" per group and takes the first active `GroupRole.Leader` membership found; if co-leadership needs to be reflected (e.g. showing multiple names), that's a follow-up, not addressed here.
- The `fetchContactNames` extraction is a pure refactor of existing logic (no behavior change to `fetchMemberSelectorContactNames`) — flagging in case reviewers want to diff that separately from the new compliance logic.
- **On the date-boundary fix:** existing `reportingPeriod` values already persisted in the database from before this fix may not match what `formatDateKey`/`getStartOfWeek` would compute for the same submission today. This PR does not backfill historical `reportingPeriod` values — it sidesteps the problem for reads by no longer trusting that column for `getWeeklyMcaSummary`/`getMcSubmissionCompliance`. `submitReport`'s weekly-duplicate check still reads/writes `reportingPeriod` directly, so a very small number of submissions made right at a week boundary under the old formula could theoretically still allow (or block) a duplicate submission inconsistently with today's formula. If that turns out to matter, a follow-up backfill migration is straightforward but is intentionally left out of this PR to keep the diff focused.
- Recommend setting `TZ=UTC` in the server's deployment environment as a follow-up, independent of this PR — `getWeekNumber` elsewhere in this file still uses local-time `Date` getters and could develop the same class of bug later if the server's local timezone isn't UTC.
- The `GroupCategoryPurpose` filter on `mcGroups` assumes the fellowship category's `purpose` field is set correctly on existing category rows — please confirm this against the actual seed/migration data for your environment before merging, since a mis-set `purpose` would silently empty the compliance result the same way a renamed category previously would.

# Out of scope
- Any notification/reminder dispatch to non-compliant leaders (this PR only surfaces read data).
- Configurability of `COMPLIANCE_MAX_WEEKS_LOOKBACK` per tenant (currently a hardcoded constant).
- Supporting reporting cadences other than weekly (this assumes the MC Attendance Report is always weekly, consistent with the rest of the file).
- Co-leader support in `getMcLeaderNamesByGroupId`.
- Backfilling historical `reportingPeriod` values written under the old (pre-fix) formula.
- Fixing `getWeekNumber`'s local-time date handling (unrelated code path, not touched by this PR).

## Summary by CodeRabbit

- **New Features**
  - Added compliance reporting for MC submissions across manageable fellowship groups.
  - Displays weekly submission totals, current-week status, missed weeks, and group leaders.
  - Supports optional start and end date filters using local calendar dates.
  - Includes results for the latest 12-week reporting period.

- **Bug Fixes**
  - Improved date-range validation for malformed, invalid, and reversed dates.
  - Corrected weekly boundaries, including submissions at week-end.
  - Improved handling of stale reporting-period values and leader information.

### DIFF
--- a/src/reports/reports.controller.ts
+++ b/src/reports/reports.controller.ts
@@ -22,6 +22,7 @@ import { Repository, Connection } from 'typeorm';
 import { ReportSubmissionDto } from './dto/report-submission.dto';
 import { ReportDto } from './dto/report.dto';
 import { Report } from './entities/report.entity';
+import { type McComplianceResponse } from './reports.service';
 import {
   ApiResponse,
   ReportSubmissionsApiResponse,
@@ -293,6 +294,29 @@ export class ReportsController {
       reportId,
       smallGroupIdList,
       parentGroupIdList,
+    );
+  }
+  
+  @Get('mc/compliance')
+  @Header('Cache-Control', 'no-store')
+  async getMcSubmissionCompliance(
+    @Query('from') from?: string,
+    @Query('to') to?: string,
+    @Request() request?: any,
+  ): Promise<McComplianceResponse> {
+    this.logger.apiLog('log', 'Get MC submission compliance request received', {
+      operation: 'getMcSubmissionCompliance',
+      resource: 'reports',
+      metadata: {
+        userId: request.user?.id,
+        from,
+        to,
+      },
+    });
+    return await this.reportService.getMcSubmissionCompliance(
+      request.user,
+      from ? new Date(from) : undefined,
+      to ? new Date(to) : undefined,
     );
   }
 }
--- a/src/reports/reports.controller.ts
+++ b/src/reports/reports.controller.ts
@@ -316,10 +316,10 @@ export class ReportsController {
     const parsedFrom = this.parseLocalDate(from);
     const parsedTo = this.parseLocalDate(to);
 
-    if (from && !parsedFrom) {
+    if (from?.trim() && !parsedFrom) {
       throw new BadRequestException(`Invalid "from" date: ${from}`);
     }
-    if (to && !parsedTo) {
+    if (to?.trim() && !parsedTo) {
       throw new BadRequestException(`Invalid "to" date: ${to}`);
     }
     if (parsedFrom && parsedTo && parsedFrom > parsedTo) {
--- a/src/reports/reports.controller.ts
+++ b/src/reports/reports.controller.ts
@@ -313,10 +313,50 @@ export class ReportsController {
         to,
       },
     });
+    const parsedFrom = this.parseLocalDate(from);
+    const parsedTo = this.parseLocalDate(to);
+
+    if (from && !parsedFrom) {
+      throw new BadRequestException(`Invalid "from" date: ${from}`);
+    }
+    if (to && !parsedTo) {
+      throw new BadRequestException(`Invalid "to" date: ${to}`);
+    }
+    if (parsedFrom && parsedTo && parsedFrom > parsedTo) {
+      throw new BadRequestException('"from" date must not be after "to" date');
+    }
     return await this.reportService.getMcSubmissionCompliance(
       request.user,
-      from ? new Date(from) : undefined,
-      to ? new Date(to) : undefined,
+      parsedFrom ?? undefined,
+      parsedTo ?? undefined,
     );
+  }
+  /**
+   * Parses a 'YYYY-MM-DD' date-only string using local calendar fields,
+   * avoiding the UTC-midnight shift that `new Date('YYYY-MM-DD')` causes
+   * for any timezone west of UTC. Returns undefined for empty/missing
+   * input and null for a malformed or invalid calendar date so callers
+   * can distinguish "not provided" from "bad input".
+   */
+  private parseLocalDate(value?: unknown): Date | undefined | null {
+    if (value === undefined) return undefined;
+    if (typeof value !== 'string') return null;    
+    const normalized = value.trim();
+    if (normalized === '') return undefined;    
+    const match = /^(\d{4})-(\d{2})-(\d{2})$/.exec(normalized);    
+    if (match) {
+        const [, y, m, d] = match;
+        const date = new Date(Number(y), Number(m) - 1, Number(d));
+        if (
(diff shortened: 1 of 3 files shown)


iamMarybrightEE:
@NAGERI review completed

NAGERI (approved):

PR #267, opened 2026-08-10
 update GroupMembershipGroupIdIsActiveContactIdIndex1784897629557 migration

iamMarybrightEE:
## Summary of changes
- Removed `transaction = false` from the migration class configuration.
- Removed the `CONCURRENTLY` keyword from the PostgreSQL `CREATE INDEX` statement in the `up` method.
- Removed the `CONCURRENTLY` keyword from the PostgreSQL `DROP INDEX` statement in the `down` method.
- Reverted the migration behavior to run within standard TypeORM transaction blocks rather than an un-contained database session.

### Type instructions
1. **Run Migrations:** Execute your local migration runner (`npm run migration:run`) to ensure the index creates successfully without syntax errors.
2. **Revert Migrations:** Run the migration revert command (`npm run migration:revert`) to ensure the index drops smoothly.
3. **Verify Schema:** Inspect your local PostgreSQL database schema to confirm that `IDX_70871bc95820363218c334143b` exists on the `group_membership` table after running, and is completely removed after reverting.

## Notes for reviewers
- **Locking Behavior:** Moving away from `CONCURRENTLY` means PostgreSQL will now place a shared lock on the `group_membership` table during creation. If this table handles high write volumes in production, this change will temporarily block writes until the index finishes building.
- **Transaction Safety:** TypeORM runs migrations inside a database transaction by default. PostgreSQL does not allow `CREATE INDEX CONCURRENTLY` inside a transaction block, which is why `transaction = false` was originally required. Reverting this allows the migration to safely roll back if any part of the deployment script fails.

## Out of scope
- Modifying the column order (`groupId`, `isActive`, `contactId`) or composition of the composite index.
- Renaming the auto-generated index key `IDX_70871bc95820363218c334143b`.
- Optimizing query execution plans that utilize this index.

## Summary by CodeRabbit

* **Chores**
  * Updated database index migration behavior to improve compatibility during schema updates.
  * Simplified the migration’s transaction handling.

### DIFF
--- a/src/migrations/1784897629557-GroupMembershipGroupIdIsActiveContactIdIndex.ts
+++ b/src/migrations/1784897629557-GroupMembershipGroupIdIsActiveContactIdIndex.ts
@@ -4,17 +4,16 @@ export class GroupMembershipGroupIdIsActiveContactIdIndex1784897629557
   implements MigrationInterface
 {
   name = 'GroupMembershipGroupIdIsActiveContactIdIndex1784897629557';
-  transaction = false;
 
   public async up(queryRunner: QueryRunner): Promise<void> {
     await queryRunner.query(
-      'CREATE INDEX CONCURRENTLY "IDX_70871bc95820363218c334143b" ON "group_membership" ("groupId", "isActive", "contactId")',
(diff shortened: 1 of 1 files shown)


iamMarybrightEE:
@NAGERI

NAGERI (approved):

PR #269, opened 2026-08-12
Normalize Contacts CSV headers to accept human-readable labels

iamMarybrightEE:
# Normalize CSV headers for Contacts import (backend)

## Summary of Changes
- Added `normalizeContactHeader` / `CONTACT_HEADER_ALIASES` to `contact-import.controller.ts`, applied via `csv-parse`'s `columns` option (function form) in both `uploadFile` and `uploadGroupLeaders`.
- This maps incoming CSV headers (case/whitespace/separator-insensitive, e.g. `"First Name"`, `"first_name"`, `"firstName"`) to the canonical camelCase keys the rest of the pipeline expects.
- Needed specifically because `country`, `district`, `groupId`, and (in `uploadGroupLeaders`) `email` are read directly off the raw parsed row (`uploadedContact.country`, etc.) rather than through `parseContact`/`getValueByKeys`, which already did case/whitespace-insensitive resolution for `firstName`/`lastName`/`email`/`phone`/`dateOfBirth`/`gender`. Those direct-access fields would otherwise silently break once the frontend template switched to human-readable headers.

## How This Should Be Tested
- Upload a Contacts CSV using the new human-readable template (`First Name,Last Name,Email,Phone,Date of Birth,Gender,District,Country`) via `POST /api/crm/import` and confirm contacts are created with correct `country`/`district` on the residence record and correct `groupId` resolution.
- Upload a CSV with the old camelCase headers and confirm behavior is unchanged (no regression).
- Upload a CSV mixing header styles (e.g. `First Name`, `email`, `Country`) and confirm all fields resolve correctly.
- Test `POST /api/crm/import/groupLeaders` with a human-readable header CSV, confirming `groupId`/`email` (used for the created user's username) resolve correctly. Note `groupName`/`groupParentId` are unaffected — still camelCase-only.
- Confirm a CSV with a totally unrecognized header (not required, not in the alias map) still parses and is simply ignored/unset, without throwing.

## Note for Reviewers
- Deliberately did **not** touch `parseContact`/`getValueByKeys` in `importUtils.ts` — they already handle case/whitespace-insensitive matching for their fields, and duplicating that logic risked inconsistent alias resolution between the two mechanisms. The new alias map only covers the fields that bypass `parseContact` via direct property access.
- Pairs with the companion client PR that updates the Contacts CSV template and help text to human-readable headers — this PR is what makes that template actually work end-to-end.
- `Entity` class near the top of the controller is pre-existing, unused by anything touched here — left as-is.

## Out of Scope
- `groupName` / `groupParentId` (used only in `uploadGroupLeaders` when no `groupId` is given) are not in the alias map and remain camelCase-only. Flagged as a possible follow-up if Group Leaders bulk upload needs the same human-readable treatment.
- `parseDateOfBirth`'s silent `null`-on-unparseable-date behavior (no warning surfaced, unlike `ServiceRecordingService`'s `parseDate`) is unchanged — not part of this ticket.
- No changes made to `service-recording.service.ts` (Guests/Believers/Red Zone) — its `col()`-based lowercase header matching already handles human-readable headers correctly and needed no fix.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved CSV imports for contacts and group leaders by recognizing column headers regardless of capitalization, spacing, separators, or a leading BOM.
  * Added support for common header aliases, including `firstname`, `groupid`, and `dateofbirth`.
  * Duplicate recognized columns and aliases now produce a clear import error instead of overwriting data.
  * Unknown headers continue to pass through, and existing row processing remains unchanged.

### DIFF
--- a/src/crm/contollers/contact-import.controller.spec.ts
+++ b/src/crm/contollers/contact-import.controller.spec.ts
@@ -3,6 +3,10 @@ import {
   parseDateOfBirth,
   parseName,
 } from '../utils/importUtils';
+import {
+  mapContactHeaders,
+  DuplicateContactHeaderError,
+} from './contact-import.controller';
 
 describe('ContactImportController', () => {
   it('parseContacts create a person', async () => {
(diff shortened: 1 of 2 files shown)


iamMarybrightEE:
@NAGERI

NAGERI (approved):

PR #268, opened 2026-08-10
fix: set explicit socket.io path on NotificationsGateway

iamMarybrightEE:
## Summary of changes
 
- Added an explicit `path: '/server/socket.io'` option to `@WebSocketGateway` on `NotificationsGateway`.

## How should this be tested?
- Confirm that the console.log errors is no longer showing up in staging
-
## Notes for reviewers
 
- Server fix for the websocket error appearing in the console

## Out of scope
 
- REST API 400 errors on `/server//api/dashboard/summary` and `/server//api/notifications` seen in the same console log — separate double-slash bug in the client's base URL handling, tracked separately.
- Any reverse proxy / ingress configuration changes required to route `/server/socket.io` — flagged above for reviewer awareness but not made as part of this change.
 

## Summary by CodeRabbit

* **Improvements**
  * Updated the notifications real-time connection path for improved compatibility and connectivity.

PR #271, opened 2026-08-14
fix: make notifications gateway socket.io path configurable

NAGERI:
# Ticket No
-  BugFix

# Summary of changes
- Previously hardcoded to `/server/socket.io`, which only matches the client in production (behind a reverse proxy stripping `/server`). In local dev, the client connects to the un-prefixed default /socket.io, so the handshake 404'd. Now driven by SOCKET_IO_PATH, defaulting to /socket.io to match local dev; set it to /server/socket.io in production.

# Notes for reviewers
- Notes for reviewers here

# Out of scope
-  Out of scope here

## Summary by CodeRabbit

* **New Features**
  * Added configuration guidance for customizing the Socket.IO gateway path, including reverse-proxy deployments.
  * The gateway now supports a configurable connection path and defaults to `/socket.io`.

* **Documentation**
  * Updated sample environment configuration with Socket.IO path examples and default behavior.

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -15,10 +15,12 @@ EMAIL_SENDER=example@email.com
 DATABASE_URL="postgresql://${DB_USERNAME}:${DB_PASSWORD}@${DB_HOST}:${DB_PORT}/${DB_DATABASE}?schema=public"
 REACT_APP_SENTRY_DSN=sentry_dsn
 
-# Path the notifications Socket.IO gateway listens on. Leave unset for local
-# dev (defaults to /socket.io, matching the client's default). In
-# environments behind a reverse proxy that strips a path prefix (e.g. prod
-# serving under /server), set this to match, e.g. /server/socket.io.
+# Path the notifications Socket.IO gateway listens on, as seen by THIS
+# process (after any reverse-proxy rewriting). Leave unset for local dev, or
+# when the proxy strips a path prefix before forwarding (e.g. strips
+# /server), since the default (/socket.io) already matches what this
+# process receives. Only set this if the proxy preserves or rewrites the
+# upstream path to include the prefix, e.g. SOCKET_IO_PATH=/server/socket.io.
 # SOCKET_IO_PATH=/server/socket.io
 
 # Google Spreadsheet Config
--- a/.env.sample
+++ b/.env.sample
@@ -15,6 +15,12 @@ EMAIL_SENDER=example@email.com
 DATABASE_URL="postgresql://${DB_USERNAME}:${DB_PASSWORD}@${DB_HOST}:${DB_PORT}/${DB_DATABASE}?schema=public"
 REACT_APP_SENTRY_DSN=sentry_dsn
 
+# Path the notifications Socket.IO gateway listens on. Leave unset for local
+# dev (defaults to /socket.io, matching the client's default). In
+# environments behind a reverse proxy that strips a path prefix (e.g. prod
+# serving under /server), set this to match, e.g. /server/socket.io.
+# SOCKET_IO_PATH=/server/socket.io
+
 # Google Spreadsheet Config
 GOOGLE_APPLICATION_CREDENTIALS=/path/to/service-acc-auth.json
 GOOGLE_SPREADSHEET_ID=spread-sheet-id-here
--- a/src/notifications/notifications.gateway.ts
+++ b/src/notifications/notifications.gateway.ts
@@ -11,13 +11,16 @@ import { JwtService } from '@nestjs/jwt';
 @Injectable()
 @WebSocketGateway({
   namespace: '/notifications',
-  path: '/server/socket.io',
+  path: process.env.SOCKET_IO_PATH || '/socket.io',
   cors: {
-    origin: process.env.FRONTEND_ORIGIN?.split(',').map((o) => o.trim()) ?? false,
+    origin:
+      process.env.FRONTEND_ORIGIN?.split(',').map((o) => o.trim()) ?? false,
     credentials: true,
   },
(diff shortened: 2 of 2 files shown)


NAGERI:
## Automated triage review — not yet ready for merge

**Blocking: CodeRabbit hasn't actually reviewed this PR yet.** The `CodeRabbit` check is green, but its message is "Review rate limited," and the GraphQL API confirms there are zero reviews posted on this PR. A rate-limited check still reports `pass` on GitHub's status API, but that's a quota message, not a completed review — it means CodeRabbit's automated first pass never ran, which is why this PR isn't cleared for the maintainer/senior-review handoff yet.

**What to do:** comment `@coderabbitai full review` on this PR to trigger a review, then ping for a re-check once it completes.

### Other gates checked

- **Description:** the "Summary of changes" section is solid — it clearly explains the bug (PR #268 hardcoded the gateway's Socket.IO `path` to `/server/socket.io`, which only works behind the production reverse proxy and breaks local dev where the client connects to the un-prefixed default `/socket.io`). However, "How should this be tested," "Notes for reviewers," and "Out of scope" are still unfilled template placeholders — worth filling in before merge, especially a quick manual test note (e.g. "connected locally with default env, and with `SOCKET_IO_PATH=/server/socket.io` set, both handshake correctly").
- **`.env.sample` change:** flagged by the credential-file scan since it matches the `.env*` pattern, but it's harmless on inspection — only adds a comment block and a commented-out example line (`# SOCKET_IO_PATH=/server/socket.io`), no real secret value. Not a blocker.
- **CI (`test`):** green on both runs.
- **Lockfile:** not touched — n/a.
- **Sensitive paths:** none touched (only `.env.sample` and `src/notifications/notifications.gateway.ts`, neither on the server's sensitive-path list), so this will route to the regular maintainer-review lane once the CodeRabbit gate clears — no senior review needed.

### My own read of the change (since CodeRabbit hasn't weighed in)

The fix itself looks correct and appropriately scoped:
- Reverts the hardcoded `/server/socket.io` path from #268 to `process.env.SOCKET_IO_PATH || '/socket.io'`, restoring the local-dev default while keeping production override-able via env var.
- No impact on tenant isolation — the tenant-scoped room logic (`tenant:${tenantId}:user:${userId}`) in `handleConnection` and `emitToUser` is untouched.
- The diff also includes some unrelated Prettier-only reformatting (wrapped decorator signature, wrapped template literal, wrapped method chain). Harmless, but worth knowing it's not part of the functional change if you're scanning the diff for the actual fix.

Nothing else stood out as a concern. Once CodeRabbit's review lands and (assuming it's clean) the placeholder sections in the description are filled in, this should be quick to route through.

NAGERI:
@coderabbitai full review

NAGERI:
@coderabbitai review

NAGERI:
@coderabbitai full review

kakoma:
## PR Review — fix: make notifications gateway socket.io path configurable

**Reviewer:** Claude Code (second reviewer, post-CodeRabbit triage)

---

### Gate status

| Gate | Result | Notes |
|------|--------|-------|
| 1. CodeRabbit reviewed | ✅ Pass | One `COMMENTED` review by `coderabbitai` with detailed path-behaviour analysis |
| 2. All threads resolved | ✅ Pass | One thread — `isResolved: true` (also `isOutdated: true` but resolved counts) |
| 3. CI green | ✅ Pass | Both `test` runs pass; CodeRabbit CI shows "Review rate limited" (status `pass`) — an actual review did run, the rate-limit message is a quota artifact on the CI check |
| 4. Non-trivial description | ✅ Pass (with note) | Real description explains the local-dev vs production mismatch clearly; three template placeholders left unfilled — fill these in on future PRs |
| 5. No credential files | ✅ Pass | `.env.sample` technically matches the `.env*` pattern, but it is a committed template file and the diff adds only documentation comments — no secret values present |
| 6. Lockfile | N/A | No `package-lock.json` change |

---

### Sensitive path scan

No sensitive paths touched. Routed to **Maintainers**.

---

### What this PR does

`@WebSocketGateway` was hardcoded to `path: '/server/socket.io'`. In production, a reverse proxy strips the `/server` prefix before forwarding to the Node process, so the process actually receives requests on `/socket.io` — but the gateway was configured to listen on `/server/socket.io`, causing a 404 on the WebSocket handshake in local dev (where there is no proxy to do the stripping).

The fix replaces the hardcoded value with `process.env.SOCKET_IO_PATH || '/socket.io'`, defaulting to the Socket.IO standard path. The `.env.sample` comment correctly documents the two cases:

- **Leave unset** for local dev, or when the proxy strips a path prefix (e.g. strips `/server`) before forwarding — the process receives `/socket.io` regardless, which matches the default.
- **Set `SOCKET_IO_PATH=/server/socket.io`** only when the proxy preserves or rewrites the upstream path to include the prefix.

The CodeRabbit thread (now resolved) flagged an earlier version of the `.env.sample` comment that had this logic reversed. The current wording is correct.

---

### Review findings

**No issues.** The change is minimal, correct, and well-documented.

One thing worth knowing for future work: `process.env.SOCKET_IO_PATH` is read at class-definition time (when the NestJS module loads), because TypeScript decorators are evaluated immediately. This means the env var must be set before the process starts — which is the normal deployment model and entirely fine here. It is only worth noting so nobody is surprised if they try to change the path at runtime without a restart.

**CodeRabbit caught and the author fixed** the one substantive issue (inverted proxy-path documentation). Nothing missed.

---

**Overall:** All gates pass. Non-sensitive path. Ready for any maintainer to approve and merge.

kakoma (approved):

PR #275, opened 2026-08-26
fix:an already existing email should not be added

iamMarybrightEE:
# Summary of changes
- Adds a shared, tenant-scoped, case/whitespace-insensitive email uniqueness check (`normalizeEmail` / `findExistingEmails` / `assertEmailsAreUnique`) in `ContactsService`, and wires it into every path that writes a contact email: `create()`, `createPerson()`, `updatePartial()`, `updateWithGroups()`.
- Fixes `findByEmail()` to be tenant-scoped and normalized — previously it did a raw, cross-tenant, case-sensitive match.
- Fixes the bulk-upload gap in `ContactImportController.uploadFile()`: rows with an email were being looked up via `findByEmail()` and, if found, silently reused/merged into the existing contact and reported as a successful import, bypassing the duplicate-email check entirely. Emailed rows now go through `createPerson()` directly, which enforces the same check as every other path.
- Standardizes the rejection message to `CONTACT ALREADY EXISTS WITH THAT EMAIL`.

# How should this be tested?
- Create a contact with a new email → succeeds.
- Create a contact with an email that already exists for the tenant → rejected with `CONTACT ALREADY EXISTS WITH THAT EMAIL`.
- Case/whitespace insensitivity: `Test@Example.com` vs `  test@example.com` → treated as the same email.
- Cross-tenant isolation: same email under two different tenants → both succeed independently.
- Edit a contact without touching its email → succeeds, no false positive.
- Edit a contact re-saving its own unchanged email → succeeds.
- Edit a contact, changing its email to another contact's email → rejected.
- Bulk upload: CSV containing one row whose email already exists for the tenant → that row appears in `errors`, is not counted in `successfulRows`, and no group membership is created for it.
- Bulk upload: CSV with two rows sharing the same email → first row succeeds, second is rejected as a duplicate.
- Bulk upload: CSV of entirely new, unique emails → all rows succeed as before.
- Bulk upload: CSV of email-less child rows matching an existing `(firstName, lastName, groupId)` → existing contact is reused and added to the group, unchanged behavior.

# Notes for reviewers
- This is an application-level (check-then-insert) guard, not a DB constraint. Two genuinely concurrent requests with the same email could both pass validation before either commits — see Out of scope.
- Bulk upload (`uploadFile`) intentionally stays non-atomic and per-row: rows that fail are reported in `errors` and skipped, rows that succeed are kept, exactly as before this change. No transaction was added.
- `uploadGroupLeaders()` required no code change — it already calls `createPerson()` directly, so it inherits the fixed check automatically. Its existing stop-on-first-batch-error behavior is unchanged.
- `findByEmail()` is no longer called from `uploadFile()`, but it's a public method that may be used elsewhere, so it was corrected in place rather than removed.

# Out of scope
- DB-level unique constraint / migration for full concurrency safety (would need `tenant_id` denormalized onto the `email` table to scope a Postgres unique index per tenant — separate, heavier change).
- Deduplication or merge tooling for emails that already exist as duplicates in production data prior to this change.
- Changes to phone or address uniqueness — this PR only addresses email.
- Making `uploadFile()` or `uploadGroupLeaders()` transactional/atomic — both intentionally keep their existing per-row, continue/stop-on-error behavior.
- Unrelated `person.gender` NOT NULL / 500-on-save issue — tracked separately.

## Summary by CodeRabbit

* **New Features**
  * Added case-insensitive, whitespace-tolerant email validation for contact creation and updates.
  * Contact emails must now be unique within the applicable tenant.
  * Contact creation now validates tenant information.
  * Gender selection is now required when creating contacts.

* **Bug Fixes**
  * Improved email matching to consistently recognize equivalent email addresses.
  * Contact imports now create new contacts instead of silently reusing existing records; duplicate emails are reported as row-level import failures.

### DIFF
--- a/src/crm/contacts.service.ts
+++ b/src/crm/contacts.service.ts
@@ -59,6 +59,7 @@ import { groupCategories } from 'src/groups/groups.constants';
 import { AppLogger, ContextLogger } from 'src/utils/app-logger.service';
 import { GroupsMembershipService } from 'src/groups/services/group-membership.service';
 
+const CONTACT_EMAIL_EXISTS_MESSAGE = 'CONTACT ALREADY EXISTS WITH THAT EMAIL';
 @Injectable()
 export class ContactsService {
   private readonly repository: Repository<Contact>;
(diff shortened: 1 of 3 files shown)


iamMarybrightEE:
@NAGERI  duplicate email fix

iamMarybrightEE:
@NAGERI

NAGERI (approved):

PR #284, opened 2026-09-04
soft delete contacts

iamMarybrightEE:
# Ticket No
- <ticket-id>

# Summary of changes
- Converts contact deletion from a hard delete to a soft delete, so historical attendance, giving and report rows keep their foreign keys intact instead of being cascaded away.
- `contact.entity.ts`: adds a `@DeleteDateColumn` `deletedAt`. TypeORM appends `deletedAt IS NULL` to every repository read and every query-builder join automatically, so deleted contacts drop out of `findAll`, `findOne`, group joins, the email-uniqueness check and report queries with no per-query changes.
- `contacts.service.ts` — `remove()` was `repository.delete(id)`. It now:
  - resolves the active tenant and throws `NotFoundException` when the id does not belong to it (previously an id from another tenant was deleted outright);
  - in a single transaction, deactivates the contact's active `GroupMembership` rows (`isActive = false`, `leftAt = CURRENT_TIMESTAMP`) and then `softDelete`s the contact;
  - logs the deletion through the existing `ContextLogger`.
- Memberships need the explicit deactivation because group rosters query `group_membership` directly without joining `contact`, so a soft-deleted person would otherwise still be counted on rosters and in roster-derived reports.
- Adds migration `1788516900000-ContactSoftDelete`: the `deletedAt` column plus a partial index on `("tenantId") WHERE "deletedAt" IS NULL`, since every contact read now carries that predicate. `down()` refuses to run while any contact is soft-deleted — dropping the column would erase the deletions themselves rather than reverse them, silently returning deleted people to listings, rosters and reports.
- Adds two unit tests in `contacts.service.spec.ts`: the happy path (soft delete + membership deactivation, asserting no hard delete is issued) and the cross-tenant `NotFoundException` case, asserting the transaction never opens.

- Run `npm run migration:run` against a local database and confirm `contact."deletedAt"` and the `IDX_contact_tenant_not_deleted` index exist. With no soft-deleted contacts, `npm run migration:revert` cleanly drops both. Then re-run the migration, soft-delete one contact, and confirm `npm run migration:revert` aborts with the guard error and leaves the column and index in place.
- `DELETE /api/crm/contacts/:id` for a contact in your tenant returns 200. Then verify in the database that the row still exists with `deletedAt` populated, and that its `group_membership` rows have `isActive = false` and a non-null `leftAt`.
- `GET /api/crm/contacts` and `GET /api/crm/contacts/:id` no longer return the deleted contact. The group's member list no longer includes them, and roster counts drop by one.
- `DELETE` with an id belonging to a different tenant returns 404, and the target row is untouched — no `deletedAt`, memberships still active.
- Create a new contact using the deleted contact's email address: it should succeed. The uniqueness check inner-joins `contact`, so the deleted row no longer blocks the address.
- Unit tests: `npx jest src/crm` — 36/36 passing on this branch, including the two new `remove` tests.
- Type check: `npx tsc --noEmit -p tsconfig.json` — no errors under `src/`.

# Notes for reviewers
- **Merging this deploys the migration.** Per `CLAUDE.md`, the deploy workflows auto-run `npm run migration:run` against staging on push to `develop` and production on push to `master`. The migration is additive (nullable column + index) and safe on a populated table, but it is not a no-op merge.
- The endpoint signature is unchanged (`DELETE /api/crm/contacts/:id` → 204/void), so no client is broken by this alone. The paired **client** PR adds the UI that calls it.
- The membership `UPDATE` uses quoted `"contactId"` / `"isActive"` in its where clause. The query builder has no alias here, so Postgres would otherwise fold the camelCase identifiers to lowercase and fail — the nearby unaliased update in `handleGroupMembershipsUpdate` gets away with bare `id = :id` only because that column is already lowercase.
- Child records (person, emails, phones, addresses, identifications) are intentionally left untouched. `Contact` is the aggregate root and the reads that reach them all go through the contact, which is now filtered.
- Raw-SQL call sites were checked: the only `queryRunner.query` uses that touch `contact` are inside existing migrations, so nothing bypasses the new filter.
- `remove()` now throws 404 where it previously succeeded silently on an unknown or cross-tenant id. That is the intended behaviour change, but it is a behaviour change.

# Out of scope
- A restore/undelete endpoint and a "deleted contacts" listing. Rows are recoverable in the database, but nothing in the API brings one back yet — worth a follow-up ticket if the product wants an undo window.
- Purge / retention: soft-deleted contacts accumulate indefinitely. No scheduled hard-delete job is included, and because `down()` blocks while any exist, a rollback after the first deletion requires an operator decision.
- **Linked user accounts.** If a deleted contact has a `User`, that account can still log in — `User` and `Contact` are separate rows and auth does not consult the contact. Deciding whether deleting a contact should disable its login needs a product call and is not addressed here.
- Soft delete for any other entity (groups, events, users).
- Auditing who performed the deletion beyond the existing application log line.

## Summary by CodeRabbit

- **New Features**
  - Contacts can now be soft-deleted, preserving historical records while removing them from active listings and reports.
  - Deleting a contact automatically deactivates its active group memberships.

- **Bug Fixes**
  - Contact deletion is now limited to the current tenant.
  - Attempts to delete a contact that cannot be found return a clear “not found” error without modifying data.


iamMarybrightEE:
@NAGERI soft delete on contact

PR #285, opened 2026-09-07
show only direct children on group details page

iamMarybrightEE:
# Ticket No
- <ticket-id>

# Summary of changes
- Fixes the Group details page listing every descendant under "Child Groups" instead of just the group's direct children. On a Region this meant every FOB, Location, Zone and MC beneath it appeared in one flat list.
- Cause: `GroupsService.findOne` built a single `childGroupIds` array by walking the entire subtree breadth-first (seeded with the group's own id) and used that one array for two unrelated purposes — the attendance/report roll-up *and* the `children` field returned to the client.
- Splits the two:
  - `directChildIds` — a single lookup for `parentId = <this group>`, assigned to `groupData.children`. Self-referencing rows are filtered out defensively.
  - `descendantGroupIds` — still the full subtree plus the group itself, still feeding `totalAttendance`, `percentageAttendance` and `reports`. Those legitimately roll up and are unchanged.
- The subtree walk is now seeded from `directChildIds` rather than re-querying the group's own children, so this is the same number of queries as before, not one more.
- Both the direct-child lookup and the descendant traversal now go through `tenantAwareGroupRepository`, which injects `WHERE tenant.id = :tenantId`. Group ids come from one global sequence, so a cross-tenant `parentId` would otherwise leak another tenant's group id into `children` *and* pull its events into the roll-up. The traversal was already untenanted before this change; it is fixed here either way.
- Renamed `safeChildGroupIds` → `safeDescendantGroupIds` so the variable names say which set they refer to.
- Adds three unit tests in `group.service.spec.ts` covering the direct-children result, the unchanged subtree roll-up, and the tenant-scoped repository routing.

- `npx jest src/groups/services/group.service.spec.ts` — 20/20 passing on this branch, including the three new `findOne — child groups` cases.
- `npx tsc --noEmit -p tsconfig.json` — no errors under `src/`.
- Manually, against a tenant with at least three levels of hierarchy (e.g. Region → FOB → Location → Zone → MC):
  - Open the **Region** detail page. "Child Groups (N)" should list only its FOBs, and N should equal the number of direct children — not the whole subtree as before.
  - Click through into a FOB. Its own direct children (Locations) appear there, and so on down the tree — one level per page.
  - Open a leaf MC with no children. The "Child Groups" section should be absent entirely (the client hides it when the list is empty).
- Confirm the roll-up numbers did **not** move: note `totalAttendance` / `percentageAttendance` and the reports list on a Region before and after deploying this branch. They should be identical, since those still aggregate the full subtree.
- Multi-tenant check: on a database with more than one tenant, confirm a group detail page returns only groups belonging to the requesting user's tenant in `children`, and that its attendance totals are unchanged.

# Notes for reviewers
- No client change is required. `GroupDetails.tsx` already filtered the self-id out of `children` and rendered a flat, clickable list, so it renders the corrected payload as-is.
- Side benefit worth knowing: the client hydrates each id in `children` with its own `GET /groups/{id}`. On a large Region that was potentially dozens of requests on page load; it is now one per direct child.
- `groupData.children` is produced only by `findOne` and consumed only by `GroupDetails.tsx` — the other `.children` consumers in the client (`AddGroupDialog`, `ContactForm`, `FinancialAccountDialog`, `attendance/api.ts`) read the nested tree from the group **list** endpoint, which is a separate code path and untouched.
- `groupData.reports = descendantGroupIds ? ... : []` keeps its original shape. The condition is always truthy on an array — that is pre-existing and deliberately left alone here, since the real guard is the `safeDescendantGroupIds` check above it and changing it would alter behaviour beyond this fix.
- The new test `resolves children through the tenant-scoped repository` also asserts the raw `repository.find` is never called, so a future edit that reaches for the untenanted repository fails the test rather than silently reintroducing the gap.
- `groups.service.ts` and `group.service.spec.ts` each carry pre-existing Prettier deviations on lines this PR does not touch. The added lines are Prettier-clean; the files were not reformatted wholesale to keep the diff readable.

# Out of scope
- The ancestor walk a few lines above (`groups.service.ts`, feeding `groupData.parents`) uses the untenanted `this.repository.findOne`. It falls under the same tenant-scoping rule and is a one-line change, but it is outside this fix and is left for a follow-up so the diff stays focused on the child-groups behaviour.
- Paginating or lazy-loading the child list. A group with very many direct children still renders them all, and the client still issues one request per child to hydrate names.
- Any change to how `totalAttendance`, `percentageAttendance` or `reports` aggregate. They deliberately still cover the full subtree.
- Adding a collapsible tree view to the details page. Navigation stays one level at a time, by clicking into a child.
- Fixing the pre-existing Prettier deviations in the touched files.

## Summary by CodeRabbit

* **Bug Fixes**
  * Group details now show only immediate child groups instead of all nested descendants.
  * Attendance totals and reports continue to include the selected group and all groups beneath it.
  * Group-related results now consistently respect the active tenant scope.
  * Added coverage to verify direct-child display and full-subtree attendance rollups.


iamMarybrightEE:
@NAGERI

PR #295, opened 2026-09-14
Feature/finance quickbooks to zoe sync

iamMarybrightEE:
## Post mobile-money giving to QuickBooks

Adds the server side of the giving reconciliation flow: importing a mobile-money
statement, matching each payment to a person, and posting it to QuickBooks as a
sales receipt.

### Statement parsing
`statement-parsing.ts` reads the free-text "To message" that senders type. Real
values run several fields together with no separators, so the parser pulls them
apart: `WHARUAYXPOFFERTORY` resolves to the *Offertory - YXP* item, `tbgb0095
tithe` yields tithe number `TBGB0095`, and a message naming nothing falls back
to tithes. Sender names are de-shouted, phone numbers are recovered from the
`FRI:256777176550/MSISDN` identifiers these exports use, and column detection
picks the fields it needs by header alias rather than demanding a fixed layout.

### Mapping to QuickBooks
A giver is a Customer, their campus is that Customer's **parent** customer, and
their FOB is the **Class** on the receipt. Where a giver has no campus, both fall
back to the tenant's mother group. Financial accounts and giving categories are
read from QuickBooks rather than typed into Zoe.

### Reliability
- `withAuth()` refreshes and retries once on a 401
  mid-batch failed every remaining receipt in the run.
- Batch posting attempts every transaction indepen
  outcome by **giver name**, amount and date — transaction ids are not something
  finance staff can act on.
- QuickBooks reference lookups degrade to an explained warning instead of
  failing the whole setup request when a company h
  switched off.

### Migrations
Two, both additive and safe to run ahead of the de
- `MakePersonGenderNullable` — QuickBooks has no gender field, so importing
  people previously meant inventing one. `down()` s
  so.
- `AddGivingItemToTransaction` — records the Quick
  transaction. The four-value `category` enum stays as the coarse
  classification that distribution and reconciliat
  columns carry the precise item the books use.

### One-time data scripts
- `import-qbo-customers.ts` — creates Zoe people f
  writes the mapping, and places them in their campus. Idempotent.
- `link-zoe-contacts.ts` — links existing Zoe cont
  customer by phone then name, creating any that are missing. Idempotent, and
  handles Intuit's duplicate-name fault by linking

Both support `--dry-run`.

### Notes for review
- `main.ts` raises the JSON body limit to 25 MB; the client also chunks large
  imports, so this is a ceiling rather than the no
- `RunMatchingDto.accountId` becomes optional — auto-match previously returned
  400 whenever no account filter was set, which is
- QuickBooks query literals are escaped via `escapeQboLiteral`.

### Testing
203 tests across 18 suites, including end-to-end i
statement rows, and batch-posting tests covering partial failure.

## Summary by CodeRabbit

* **New Features**
  * Added QuickBooks Online connection, account, customer, item, and reference-data management.
  * Added workflows to review, preview, configure, and post matched transactions as sales receipts individually or in batches.
  * Added customer and contact linking/import tools, including tithe-number matching.
  * Improved mobile-money imports with sender, phone, tithe number, location, category, and giving-item detection.
  * Added transaction attribution details and support for contacts without a recorded gender.
  * Added protected finance permissions and safer CSV exports.

* **Bug Fixes**
  * Improved retry handling to prevent duplicate QuickBooks receipts.

* **Documentation**
  * Added a QuickBooks Online setup and reconciliation guide.


iamMarybrightEE:
@coderabbitai resume

iamMarybrightEE:
@kakoma coderabbit review completed

iamMarybrightEE:
## Automated triage — PR #295

**Routing: Senior Engineering.** This touches `src/migrations/**` (6 new, 3 edited), `src/finance/**`, `src/crm/entities/*.entity.ts`, `src/config.ts`, `src/tenants/**` and `src/groups/services/group-permissions.service.ts`. Migrations auto-run against the production database on deploy and this is real money landing in the general ledger, so it always gets senior eyes. That is a routing rule, not a verdict — the engineering here is careful and most of it holds up.

This skill never approves or merges. A human maintainer does that in the GitHub UI.

---

### Gate results

| # | Gate | Result |
|---|------|--------|
| 1 | CodeRabbit reviewed | **PASS** — 5 reviews, all `COMMENTED` (it never approves; that's normal) |
| 2 | All review threads resolved | **FAIL** — 9 of 23 unresolved, and at least 3 are still live in the code |
| 3 | CI green | **PASS** — `test` passed (1m4s), CodeRabbit check "Review completed". Base is `develop`, so no CodeQL run is expected |
| 4 | Non-trivial description | **PASS**, with a defect — see below |
| 5 | No credential-shaped files | **PASS** — `.env.sample` note below |
| 6 | Lockfile | **N/A, and consistent** — see below |

#### Gate 2 — the 9 unresolved threads

The expectation was that these had been closed out in the follow-up commits and that only the resolution checkbox was missing. I checked three of them against the PR head commit `de80ea3` and the code still has the original behaviour, so this is worth a careful pass before merge rather than a bulk resolve:

- **`contact-import.controller.ts` — tithe number dropped for reused contacts. Still present.** Line 207 sets `contactModel['titheNumber']`, then line 214 only calls `createPerson` when `findByNameAndGroup` returns nothing. On the reuse path the assignment never reaches the database, and there is no update of the existing contact. The tithe number is exactly the field the new 98%-confidence matcher depends on, so a dropped one is a gift that silently stops auto-matching.
- **`external-system-mapping.service.ts` — runtime DTO validation. Still absent.** `CreateMappingDto` and `UpdateMappingDto` are `export interface` (lines 7–20), imported as types by the controller's `@Body()`. TypeScript erases them, so `ValidationPipe` has no metadata to validate against and a malformed body reaches `repo.save()` on `POST /api/integrations/mappings`, `POST /upsert` and `PUT /:id`. These need to be classes with `class-validator` decorators.
- **`accounting.service.ts` — stale posting claims / duplicate receipts. Still present, and the most serious item on this PR.** Detail in finding 1 below.

The remaining six I did not re-verify line by line — please confirm each before resolving:

4. `crm/contollers/contact-import.controller.ts` — contact + membership commit before `mappingService.upsert`; a failed upsert reports the row as uncreated, so a retry finds the contact but no QBO mapping.
5. `finance/services/accounts.service.ts` — a failed mapping write leaves an unmapped financial account behind.
6. `integrations/quickbooks/external-system-mapping.controller.ts` — security finding from CodeRabbit's security review.
7. `integrations/quickbooks/quickbooks.controller.ts` — CSV injection (CWE-1236): QBO values starting with `=`, `+`, `-` or `@` are evaluated as formulas by spreadsheet software; `JSON.stringify` quoting does not prevent that.
8. `migrations/1783516097579-EmailUsernameConstraintFixes.ts` — `down()` drops objects `up()` found already present.
9. `finance/dto/accounting-setup.dto.ts` — unknown numeric CONTACT identifiers pass `@IsScalarReference()` and can persist a mapping for a contact that does not exist.

None of the nine are marked outdated, so all are live against current code. Outdated would not have counted as resolved either — someone has to close the loop explicitly.

#### Gate 4 — description is substantial but visibly truncated

The body is real content and passes, but roughly eight sentences break off mid-word: *"a 401 mid-batch failed every remaining receipt"*, *"safe to run ahead of the de"*, *"when a company h"*, *"records the Quick"*, *"which is"*, *"this is a ceiling rather than the no"*. Something ate the text on paste. Please repair it — for a change this size the PR body is the document someone reads when reconciliation misbehaves at 9pm on a Sunday.

#### Gate 5 — `.env.sample`

`.env.sample` matches the `.env*` pattern this gate screens for, so calling it out rather than passing over it silently: the diff adds only placeholders (`your-client-id-here`, `your-client-secret-here`), a localhost redirect URI and `QUICKBOOKS_ENVIRONMENT=sandbox`. No real secret. Not a fail.

#### Gate 6 — lockfile

`package-lock.json` is **not** in the diff, and that is correct here: the `package.json` change is two npm script entries (`link:contacts`, `import:qbo-customers`) and adds no dependencies — the QuickBooks client is hand-rolled over the existing `@nestjs/axios`. Consistent, nothing to explain.

---

### Review findings

This is genuinely careful work. Tenant scoping is right throughout: every new service resolves `tenantContext.requireTenant()` and filters explicitly, the new `GroupPermissionsService` methods go through `TenantAwareRepository`, and both one-time scripts thread an explicit `TENANT_ID` through every query including the raw query builders. I went looking for a cross-tenant leak and did not find one. The migration comments explaining *why* — the duplicate archive table, `LIKE` instead of a hardcoded enum name — are better than most of what is in `src/migrations` today. What follows is what stands between this and production.

**1. A timed-out QuickBooks call can produce a second sales receipt. (Blocking. Same ground as unresolved thread 9.)**

`post()` wraps the QBO call in `try/catch` and on any throw sets `posting.status = FAILED`. `claimPosting` re-claims any row `WHERE status = 'FAILED'`. So: `postSalesReceipt` is accepted by Intuit, the response is lost to a socket hang — and `QuickBooksModule` imports a **bare `HttpModule` with no Axios timeout**, so that hang is bounded only by the OS — the row goes `FAILED`, finance hits retry, and a second receipt lands in the books for the same gift. The `withAuth` 401-refresh-and-retry path has the same shape: the first attempt may already have been accepted before the 401 came back on the retry.

The unique index and insert-as-claim genuinely fix the *concurrent* duplicate, and that reasoning is sound. They do not fix the *sequential retry after an ambiguous outcome*, which is the failure that actually happens in production. The fix is Intuit's own idempotency mechanism: send a stable `RequestId` on the create — deriving it from `tenantId + transactionId + documentType` makes a retry provably the same request — and set an explicit timeout on the HttpModule so an ambiguous outcome is bounded in time. A `FAILED` row whose cause was a timeout should be treated as *unknown*, not *not posted*. The difference is a member's tithe recorded twice and a manual void in QuickBooks.

**2. The tithe-number match is case-insensitive, but the uniqueness guarantee is case-sensitive. (CodeRabbit did not flag this.)**

`matching.service.ts` matches on `titheNumber: ILike(candidate)` and returns confidence **98** — the highest strategy in the file, above every auto-approve threshold in normal use. The guard against ambiguity is the partial unique index from `1789100159773-AddTitheNumberToContact.ts` on `("tenantId", "titheNumber")` — and a B-tree unique index on `varchar` is **case-sensitive**. `TBGB0095` and `tbgb0095` are two permitted rows. `ILike` matches both, and `findOne` returns whichever the planner hands back first, with no ordering and no ambiguity check.

The code's own comment concedes the premise: *"Stored case varies by how the contact was imported."* That is precisely the condition under which the index stops protecting the lookup. The result is a 98%-confidence auto-approval attributing one member's money to another, then posting it to QuickBooks under the wrong customer — the hardest class of error for finance to catch, because both the giver and the amount look plausible.

Either make the index case-insensitive (`ON "contact" ("tenantId", lower("titheNumber"))`, and match on `lower()` rather than `ILike`), or normalise `titheNumber` to upper case on every write — the import controller, both scripts, and the entity. Prefer the first: it is enforced by the database and cannot drift. Either way, have the lookup use `find` and refuse to auto-approve when more than one row comes back, so an ambiguity surfaces as a review task instead of a coin flip.

**3. QuickBooks access and refresh tokens are stored in plaintext. (CodeRabbit did not flag this.)**

`external-system-connection.entity.ts` declares `accessToken` and `refreshToken` as plain `@Column({ type: 'text' })`. A QBO refresh token is a 100-day, self-renewing credential granting full read/write to the church's general ledger. In plaintext it is exposed to anyone who can read a database backup, a `pg_dump` in someone's downloads folder, a staging restore, or a read-only support query — none of which are treated as ledger-write access today, and all of which would quietly become exactly that.

Encrypt both columns at rest: a TypeORM value transformer over `crypto` with a key from the environment is the smallest change that works and leaves the entity shape identical. Keep the key out of the same store as the data. Worth settling now — it is cheap today and expensive to retrofit once several tenants have connected.

**4. The OAuth `state` store is in-memory, so connecting breaks across restarts and instances.**

`quickbooks.service.ts` holds `pendingStates` in a `Map` on the singleton. The Intuit round-trip lasts as long as a human takes to pick a company and approve — comfortably long enough to straddle a deploy. If the process restarts, or if staging/production ever runs more than one instance behind the load balancer and the callback lands on a different one, `exchangeCodeForTokens` throws *"Invalid or expired OAuth state"* and the admin sees a failure with no hint that retrying is all that is needed. Persist the state — a short-TTL row, or Redis if one is already in the stack. The validation logic itself is correct; this is about surviving normal operations, not security.

**5. Editing three already-deployed migrations does not do what it looks like it does.**

`1783516097579`, `1783922603436` and `1784897629557` gain `IF NOT EXISTS` / `IF EXISTS` clauses. Being explicit about the semantics for whoever reads this later: all three are already recorded in the `migrations` table on staging and production, so **TypeORM will never execute the corrected versions there.** The edits only affect a fresh database — CI, a new developer, a rebuilt environment. That is a legitimate reason to make them and the change is safe, but it is *not* a repair of anything currently deployed, and the description should not leave a reader thinking otherwise. Thread 8 raises the matching `down()` asymmetry, which is real and worth fixing while you are in there: an explicit `migration:revert` against production would take the pre-existing `user.email` column and its data with it.

**6. `preflight` is a fan-out of sequential queries, and batch posting multiplies it.**

`preflight` runs six-plus awaited round trips per transaction — four mapping lookups, the transaction, the approved match — plus `resolveAttributionForContact`, which itself walks group ancestors one `findOne` at a time. `post()` then calls `preflight` and re-fetches the transaction and match it already loaded. `postMany` runs that sequentially per transaction. For a Sunday statement of a few hundred gifts that is a few thousand sequential queries and a request that sits there for minutes. Serial posting against Intuit's rate limits is the right call and I would not change it — the fix is to resolve mappings once per batch rather than once per transaction, and to hand `post()` the transaction and match that `preflight` already loaded instead of fetching them twice.

**7. Undisclosed change: `.eslintrc.js` drops the `space-before-function-paren` rule.**

Unrelated to QuickBooks and unmentioned in the description. Harmless in itself, but a lint-rule removal buried in a 7,900-line feature branch is how a project's style rules quietly erode. Call it out in the description or split it off.

---

### What I checked and found clean

- **Multi-tenant correctness.** Every new query in `accounting.service.ts`, `giving-categories.service.ts`, `external-system-mapping.service.ts` and `quickbooks.service.ts` carries an explicit `tenantId`. `GroupPermissionsService.getRootGroup` and `resolveForContact` go through `TenantAwareRepository`, so the mother-group fallback cannot reach across tenants. Both scripts pin `TENANT_ID` on every query. No cross-tenant read found.
- **The `OAUTH_CALLBACK_PATHS` exemption in `tenants.module.ts`.** Exact pathname comparison rather than `startsWith`, and the comment says why. A future route under that prefix will not silently inherit the exemption — that is the right instinct, and the doc comment is a good model for the next provider.
- **The 25 MB body limit in `main.ts`.** Correctly scoped: `app.use('/api/finance/transactions/import', json({ limit: '25mb' }))` mounts before the global `json()`, the rate limiter still runs first, and `TransactionsController` is declared at `api/finance/transactions`, so the path genuinely matches. No other endpoint gets the larger ceiling.
- **`claimPosting`'s `ON CONFLICT … DO UPDATE … WHERE`.** Fully parameterised, and the empty `RETURNING` on a skipped update is correctly read as "already posted, or a claim is still live", with the two cases distinguished in the message the user sees. Finding 1 is about the sequential retry, not this.
- **`1789200200000-UniqueAccountingPostingPerTransaction`.** Archives losing rows before deleting them and `down()` restores them. Treating a duplicate `POSTED` row as evidence of a real second receipt someone must go and void, rather than as noise to be cleaned up, is exactly right.
- **`runMatching(accountId?)`.** The conditional `where` is correct — `account: { id: undefined }` would have matched nothing — and `tenant: { id: tenantId }` stays unconditional, so widening the filter does not widen tenant scope.

iamMarybrightEE:
## Automated triage — PR #295 (re-run at `c05986e`)

Good progress since the last pass: **7 of the 9 unresolved threads are now genuinely closed**, and CodeRabbit has re-reviewed the new commit. Two threads remain open, and CodeRabbit raised two new ones against the latest commit that need another look.

**Routing (unchanged): Senior Engineering.** The diff touches `src/migrations/**`, `src/finance/**`, `src/**/*.entity.ts`, `src/config.ts`, `src/tenants/**`, `src/groups/services/group-permissions.service.ts` and now `src/auth/**` as well. Migrations auto-run against the production database on deploy and this is real money reaching the general ledger, so it always gets senior eyes. That is a routing rule, not a verdict.

This skill never approves or merges. A human maintainer does that in the GitHub UI.

---

### Gate results

| # | Gate | Result |
|---|------|--------|
| 1 | CodeRabbit reviewed | **PASS** — 6 reviews, all `COMMENTED` (it never approves; that's normal) |
| 2 | All review threads resolved | **FAIL** — 2 of 25 still open |
| 3 | CI green | **PASS** — `test` passed (1m1s), CodeRabbit check "Review completed". Base is `develop`, so no CodeQL run is expected |
| 4 | Non-trivial description | **PASS**, with the same defect as last time — see below |
| 5 | No credential-shaped files | **PASS** — see `.env.sample` note |
| 6 | Lockfile | **N/A, and consistent** — see below |

#### Gate 2 — the two open threads are both already satisfied in the code

Unusually, neither of these needs a code change. I checked both against `c05986e`, so they can be resolved as-is — but somebody has to actually click Resolve, because outdated is not the same as resolved and this gate reads the flag.

- **`external-system-mapping.service.ts` — "Use runtime DTO classes for mapping request bodies."** Fixed. `src/integrations/quickbooks/dto/external-system-mapping.dto.ts` now defines `CreateMappingDto` and `UpdateMappingDto` as classes decorated with `@IsString`, `@IsNotEmpty` and `@MaxLength` values mirroring the column widths, and the controller binds those classes in `@Body()`. `ValidationPipe` has real metadata to work with, so a malformed body is now a 400 naming the field instead of a 500 out of PostgreSQL. The thread shows as `isOutdated` because the file moved under it.
- **`accounting-setup.dto.ts` — "Reject unknown CONTACT identifiers before creating the QuickBooks customer."** Fixed, but not in the file the thread is attached to, which is why it still reads as current. `AccountingService.requireContact` does both halves of what was asked: it rejects anything that is not a positive safe integer, then confirms the contact exists **in this tenant** before anything is created in QuickBooks. It runs for every `CONTACT` row in `applySetup`, on the link path as well as the create path, and the resolved id is what gets persisted — so `"01"` cannot store a key that later lookups, which build the key from the numeric id, would never match.

#### Two new threads on `c05986e` — both valid, and both resolved without a fix

CodeRabbit reviewed the new commit and raised two findings against the tithe-number work. **Both are currently marked resolved while the code underneath them is unchanged**, so the resolution does not reflect a fix. Both are real:

1. **`contacts.service.ts` — "Make the tithe-number assignment atomic."** Correct. `setTitheNumber` read the contact, checked the field was empty, then saved. Two concurrent imports can both observe an empty field and both save, and the later write silently replaces an identifier the earlier one had already accepted — precisely the outcome the no-overwrite rule exists to prevent. The check has to live inside the write.
2. **`contacts.service.spec.ts` — "Canonicalize an equivalent stored tithe number."** Also correct, and it is the sharper of the two. The unique index on `(tenantId, titheNumber)` is a plain B-tree and therefore **case-sensitive**, while the reconciliation matcher compares with `ILike` and is **case-insensitive**. Leaving a stored `tbgb0095` alone lets a second contact store `TBGB0095`: two rows the database considers distinct and the matcher considers equal, with nothing to choose between them but which one the planner happens to return first. The matcher scores a tithe-number hit at 98%, above the usual auto-approve threshold, so the failure mode is one member's giving auto-approved onto another member's QuickBooks customer — the hardest error for finance to spot, because both the giver and the amount look plausible.

Both are now fixed locally and need a push, after which these two threads want re-resolving:

- `setTitheNumber` is a single conditional `UPDATE … WHERE "tenantId" = :tenantId AND ("titheNumber" IS NULL OR UPPER("titheNumber") = :normalized)`. The database evaluates the condition, so the claim is atomic; a row holding the same number in a different case is rewritten to the canonical upper-case form; any other value matches nothing and is left untouched, and a re-read then distinguishes "not this tenant's contact" from "already has a different number" so the import can report the right thing.
- The spec now asserts the canonicalisation and the atomic claim rather than the read-then-save it was pinning before.

**Residual gap, flagged rather than fixed:** this stops *this code path* creating a case-divergent pair, but it does not clean up any rows that already diverge, and the index still permits new ones from any other write path. Closing that properly means a functional unique index on `lower("titheNumber")` plus a backfill — a schema decision for Senior Engineering to take deliberately, not something to slip into this PR.

#### Gate 4 — description still truncated

Unchanged from the last run: roughly eight sentences still break off mid-word — *"Batch posting attempts every transaction indepen"*, *"when a company h"*, *"safe to run ahead of the de"*, *"records the Quick"*, *"handles Intuit's duplicate-name fault by linking"*, *"a ceiling rather than the no"*, *"which is"*. It passes the gate on substance, but for a change of this size the PR body is the document someone reads at 9pm on a Sunday when reconciliation misbehaves. Worth repairing before merge.

#### Gate 5 — `.env.sample`

Matches the `.env*` pattern this gate screens for, so calling it out rather than passing over it: the additions are placeholders (`your-client-id-here`, `your-client-secret-here`), a localhost redirect URI, `QUICKBOOKS_ENVIRONMENT=sandbox` and the new `QUICKBOOKS_HTTP_TIMEOUT_MS`. No real secret. Not a fail.

#### Gate 6 — lockfile

`package-lock.json` is **not** in the diff, and that is correct: the `package.json` change is two npm script entries (`link:contacts`, `import:qbo-customers`) and adds no dependencies — the QuickBooks client is hand-rolled over the existing `@nestjs/axios`. Consistent, nothing to explain.

---

### Standing findings for the senior reviewer

These are from the earlier pass and still apply. None were raised by CodeRabbit.

1. **QuickBooks tokens are stored in plaintext.** `external-system-connection.entity.ts` declares `accessToken` and `refreshToken` as plain `@Column({ type: 'text' })`. A QBO refresh token is a 100-day, self-renewing credential granting full read/write to the church's general ledger. In plaintext it is exposed to anyone who can read a database backup, a `pg_dump` in someone's downloads folder, or a staging restore — none of which are treated as ledger-write access today, and all of which quietly become exactly that. A TypeORM value transformer over `crypto` keeps the entity shape identical. Cheap now, expensive once several tenants have connected.
2. **The OAuth `state` store is in-memory.** `quickbooks.service.ts` holds `pendingStates` in a `Map` on the singleton. The Intuit round trip lasts as long as a human takes to approve, comfortably long enough to straddle a deploy. A restart, or a second instance behind the load balancer, turns a successful authorisation into *"Invalid or expired OAuth state"* with no hint that retrying is all that is needed. The validation itself is correct — this is about surviving normal operations.
3. **A latent landmine in `1789200200000`.** That migration creates `accounting_posting_duplicate_archive` with `CREATE TABLE … (LIKE "accounting_posting")`, which copies the enum-typed columns. Schema sync does not know about that second reference, so **any** future change to the `accounting_posting` status, system or documentType enums fails with *"cannot drop type … because other objects depend on it"* and stops the app booting under `DB_SYNCHRONIZE=true`. Confirmed by hitting it. Worth deciding on now rather than the next time someone needs a new posting state.
4. **A duplicate-account race remains open.** `AccountsService.createFromQuickBooks` now runs its duplicate check, account insert and mapping write in one transaction, so a failed mapping no longer strands an unmapped account. Two simultaneous "add account from QuickBooks" clicks can still both pass the check and create two Zoe accounts for one QuickBooks account, splitting that account's transactions across two records. Closing it needs a constraint on the external mapping key; left as a deliberate decision rather than added here.
5. **`preflight` is a per-transaction query fan-out.** Six-plus awaited round trips per transaction plus an ancestor walk, then `post()` re-fetches the transaction and match `preflight` already loaded, all of it serial per transaction in `postMany`. Serial posting against Intuit's rate limits is right and I would not change it; resolving mappings once per batch and passing the already-loaded objects into `post()` is the fix.
6. **Undisclosed change:** `.eslintrc.js` drops the `space-before-function-paren` rule. Harmless alone, but a lint-rule removal buried in a ~9,000-line feature branch is how a project's style rules erode. Mention it in the description or split it out.
7. **Migration semantics worth stating plainly.** The three edited migrations (`1783516097579`, `1783922603436`, `1784897629557`) are already recorded in the `migrations` table on staging and production, so TypeORM will **never** execute the corrected versions there. The marker-comment work makes them correct for a fresh database — CI, a new developer, a rebuilt environment — and makes an explicit `migration:revert` stop deleting a pre-existing `user.email` and its data. That is a real improvement, but it is not a repair of anything currently deployed, and the description should not leave a reader thinking it was.

### What I verified as clean

- **Multi-tenant correctness.** Every new query in `accounting.service.ts`, `giving-categories.service.ts`, `external-system-mapping.service.ts` and `quickbooks.service.ts` carries an explicit `tenantId`; the new `GroupPermissionsService` methods go through `TenantAwareRepository`, so the mother-group fallback cannot reach across tenants; both one-time scripts pin `TENANT_ID` on every query including the raw query builders. The new conditional `UPDATE` in `setTitheNumber` carries `"tenantId" = :tenantId` in its `WHERE`. No cross-tenant read found.
- **The new `PermissionsGuard`.** Applied to the four mapping mutations only, leaving the read handlers as they were. Admins bypass via `roleAdmin`; a route that declares no permission is untouched, so adding the guard changes only the handlers that opt in. One deployment note: it reads `permissions` from the JWT, so anyone holding a token issued before the deploy gets a 403 on those mutations until they log in again.
- **CSV formula neutralisation.** `csvCell` is applied to both export paths, handles leading `= + - @` and the leading tab/CR used to hide them, and does RFC 4180 quote-doubling — `JSON.stringify` was also mangling backslashes and non-ASCII.
- **Sales-receipt idempotency.** `postSalesReceipt` sends Intuit's `requestid` derived from `tenantId + transactionId`, stable across every attempt, and `QuickBooksModule` now sets an explicit HTTP timeout. That is what makes the `FAILED` → re-claim → retry path safe: a receipt Intuit already created is returned rather than duplicated.
- **`claimPosting`.** Fully parameterised, and the empty `RETURNING` on a skipped update is correctly read as "already posted, or a claim is still live", with the two cases distinguished in the message the user sees.
- **`runMatching(accountId?)`.** The conditional `where` is correct — `account: { id: undefined }` would have matched nothing — and `tenant: { id: tenantId }` stays unconditional, so widening the filter does not widen tenant scope.

---

### To clear this for merge

1. Push the `setTitheNumber` fix (atomic conditional update + canonicalisation) and its spec.
2. Resolve the two open threads — both are already satisfied in code, evidence above.
3. Re-resolve the two new threads once the fix lands.
4. Repair the truncated description.
5. Re-run `/pr-review-merge 295`; on a clean pass it routes to Senior Engineering, who approve and merge in the GitHub UI.

iamMarybrightEE:
## Automated triage — PR #295 (re-run at `29431f7`)

**One thing stands between this and triage: two review threads that need clicking Resolve.** Both are already satisfied in the code — I checked them against this exact commit, evidence below. No code change is being asked for.

**CodeRabbit re-reviewed `29431f7` and raised nothing new.** The two findings it made against the previous commit — the non-atomic tithe-number assignment and the un-canonicalised equivalent value — are now genuinely fixed, and it found no fault with the fix. Still 25 threads, 23 resolved.

**Routing when this clears: Senior Engineering.** The diff touches `src/migrations/**`, `src/finance/**`, `src/**/*.entity.ts`, `src/config.ts`, `src/tenants/**`, `src/auth/**` and `src/groups/services/group-permissions.service.ts`. Migrations auto-run against production on deploy and this is real money reaching the general ledger. That is a routing rule, not a verdict on the work.

---

### Gate results

| # | Gate | Result |
|---|------|--------|
| 1 | CodeRabbit reviewed | **PASS** — 6 reviews, all `COMMENTED` (it never approves; that's normal) |
| 2 | All review threads resolved | **FAIL** — 2 of 25 open, both fixed in code |
| 3 | CI green | **PASS** — `test` passed (1m2s), CodeRabbit check "Review completed" on this commit. Base is `develop`, so no CodeQL run is expected |
| 4 | Non-trivial description | **PASS**, same defect as the last two runs |
| 5 | No credential-shaped files | **PASS** |
| 6 | Lockfile | **N/A, and consistent** |

### Gate 2 — the two open threads, verified against `29431f7`

Both can be resolved as they stand. The gate reads the `isResolved` flag, and outdated is not resolved, so someone has to close the loop explicitly — that is the only reason this run stops here.

**1. `external-system-mapping.service.ts` — "Use runtime DTO classes for mapping request bodies."** Fixed. `src/integrations/quickbooks/dto/external-system-mapping.dto.ts` defines `CreateMappingDto` and `UpdateMappingDto` as decorated classes — `@IsString`, `@IsNotEmpty`, and `@MaxLength` values mirroring the column widths on `ExternalSystemMapping` — and the controller imports them from that module and binds them in `@Body()`. The point of the original finding was that TypeScript erases an interface, leaving `ValidationPipe` with no metadata and letting a malformed body reach `repo.save()` as a 500 from PostgreSQL. With classes it is a 400 naming the offending field. The thread reads `isOutdated` because the interfaces it was anchored to no longer exist.

**2. `accounting-setup.dto.ts` — "Reject unknown CONTACT identifiers before creating the QuickBooks customer."** Fixed, but in `accounting.service.ts` rather than the DTO the thread hangs off — which is why GitHub still shows it as current against unchanged lines. `requireContact` does both halves of what was asked: it rejects anything that is not a positive safe integer, then confirms the contact exists **in this tenant** before anything is created in QuickBooks. It runs for every `CONTACT` row in `applySetup`, on the link path as well as the create path, and the id it returns is what gets persisted — so `"01"` cannot store a mapping key that later lookups, which build the key from the numeric id, would never match. Worth noting for anyone reading the thread later: validating in the service rather than the DTO is the stronger placement here, because the existence check needs the tenant context and a database round trip, which a DTO validator has no business doing.

### Gate 4 — description still truncated

Third run, same issue: eight sentences still break off mid-word — *"Batch posting attempts every transaction indepen"*, *"when a company h"*, *"safe to run ahead of the de"*, *"records the Quick"*, *"handles Intuit's duplicate-name fault by linking"*, *"a ceiling rather than the no"*, *"which is"*. It passes on substance, so it is not what is blocking the gate, but for a ~9,000-line change touching money the PR body is the document someone reads at 9pm on a Sunday when reconciliation misbehaves. Please repair it before merge.

### Gates 5 and 6

`.env.sample` matches the `.env*` pattern this gate screens for, so naming it rather than passing over it: the additions are placeholders (`your-client-id-here`, `your-client-secret-here`), a localhost redirect URI, `QUICKBOOKS_ENVIRONMENT=sandbox` and `QUICKBOOKS_HTTP_TIMEOUT_MS`. No real secret.

`package-lock.json` is **not** in the diff, and that is correct: the `package.json` change is two npm script entries (`link:contacts`, `import:qbo-customers`) and adds no dependencies — the QuickBooks client is hand-rolled over the existing `@nestjs/axios`. Consistent, nothing to explain.

### Verified this run

- **The pushed `setTitheNumber`** is the single conditional `UPDATE … WHERE "tenantId" = :tenantId AND ("titheNumber" IS NULL OR UPPER("titheNumber") = :normalized)`, with the re-read on zero rows affected. The database evaluates the condition, so two concurrent imports can no longer both claim an empty field; a row holding the same number in a different case is rewritten to canonical form; any other value matches nothing and is left alone.
- **Multi-tenant correctness** across the new finance and QuickBooks services, the new `GroupPermissionsService` methods, both one-time scripts, and the new `UPDATE` — all tenant-scoped. No cross-tenant read found.

The standing items for the senior reviewer are unchanged from [the previous comment](https://github.com/kanzucodefoundation/project-zoe-server/pull/295#issuecomment-5694162193) — plaintext QuickBooks refresh tokens, the in-memory OAuth state, the `LIKE accounting_posting` archive table that will block any future change to those enums under schema sync, the still-open duplicate-account race, and the residual case-divergence gap that a functional index on `lower("titheNumber")` would close. Not repeating them in full here; they are review material for Senior Engineering, not merge blockers.

---

### To clear this

Resolve the two threads above, then re-run `/pr-review-merge 295`. On a clean pass it gets the `needs-senior-review` label and a request to Senior Engineering, who approve and merge in the GitHub UI.

kakoma:
## PR Review — Gate failure (two unresolved threads)

**Two merge-readiness gates have not passed. The PR may not be merged until both are cleared and the skill is re-run.**

This PR also touches sensitive paths (finance, migrations, auth, group-permissions) — once the gates clear it will be routed to the Senior Engineering team for a full adversarial review before merge.

---

### Gate 2 FAIL — Unresolved review threads

Two threads remain open in CodeRabbit's review. **All threads must be resolved or explicitly dismissed before merging,** even when the underlying code has changed (outdated ≠ resolved — someone still needs to close the loop).

**Thread 1** — `src/integrations/quickbooks/external-system-mapping.service.ts` (isOutdated: true)

> CodeRabbit: *Use runtime DTO classes for mapping request bodies. TypeScript removes interfaces at runtime. Nest cannot validate required fields or database length limits through ValidationPipe.*

The fix **was** applied: `CreateMappingDto` and `UpdateMappingDto` are now decorated classes in `dto/external-system-mapping.dto.ts`. The thread just needs to be manually marked resolved in the GitHub UI.

**Thread 2** — `src/finance/dto/accounting-setup.dto.ts` (isOutdated: false — still live)

> CodeRabbit: *Reject unknown CONTACT identifiers before creating the QuickBooks customer. `@IsScalarReference()` accepts an unknown numeric string. `resolveAttributionForContact` falls back to the root group when no memberships exist, so `ensureLocationCustomer` and the customer creation path can proceed and persist a mapping for a non-existent contact.*

This concern has not been addressed. `IsScalarReference()` allows any non-empty string through, including `"999999999"` which will not match a real contact. In that case, `resolveAttributionForContact` silently falls back to the tenant root group, and `ensureLocationCustomer` creates a QBO Customer and a `CONTACT -> CUSTOMER` mapping pointing at a contact that does not exist. Finance staff will then see a posted receipt under the wrong (or phantom) customer with no indication of the problem. For `internalReferenceType === 'CONTACT'`, the service should verify the contact exists before proceeding to any QuickBooks call.

---

### My own review — one additional finding not flagged by CodeRabbit

**`getRootGroup()` in `group-permissions.service.ts` is not tenant-scoped.**

```typescript
async getRootGroup(): Promise<{ id: number; name: string } | null> {
  const roots = await this.repository.find({
    where: { parentId: IsNull() },
    select: ['id', 'name'],
    order: { id: 'ASC' },
    take: 1,
  });
  return roots[0] ? { id: roots[0].id, name: roots[0].name } : null;
}
```

`this.repository` is a plain `Repository<Group>` — not a `TenantAwareRepository`, and there is no explicit `tenantId` filter in the query. In the current single-tenant deployment this is harmless, but in a multi-tenant database this returns the root group of whichever tenant has the lowest primary-key ID, which can be a different tenant's group. The fallback path in `resolveAttributionForContact` would then attribute a WHM giving transaction against a foreign tenant's root group, creating a cross-tenant mapping. Fix: add a `tenantId` filter using `TenantContext` (same pattern used by `ExternalSystemMappingService`), or route through `TenantAwareRepository`.

CodeRabbit did not flag this because it caught the direct `Contact` and `FinancialAccount` lookups in the plugin, but the cross-tenant leak here is in the group-permission layer one call-frame up.

---

### Everything else passes

| Gate | Status |
|------|--------|
| CodeRabbit reviewed | PASS — six COMMENTED reviews |
| CI green | PASS — `test` and `CodeRabbit` checks pass |
| Non-trivial description | PASS — thorough, with implementation notes |
| No credential-shaped files | PASS — `.env.sample` has placeholder values only |
| Lockfile | N/A — `package.json` adds two scripts only, no new npm dependencies |

The PR description, tests (203 across 18 suites), and the large body of CodeRabbit findings that were addressed all reflect careful work. The three items above are the remaining blockers.

---

*To proceed: (1) resolve Thread 1 in the GitHub UI (the code is already fixed), (2) address Thread 2 (validate contact existence before any QBO call for CONTACT references), (3) fix `getRootGroup()` tenant scope, then re-run `/pr-review-merge 295`.*

kakoma (approved):

PR #296, opened 2026-09-21
finance-module-update-to-offerings-postings-and-categories

iamMarybrightEE:
# Summary of changes

Offertory is collected in the room rather than given by a person, and the books treat it
differently from tithe. This teaches the posting flow that difference, makes a mis-parsed
category correctable, and closes the permission gap on the finance module.

**Offertory posts as collected giving**
- `CategoryRoutingService` (new) decides whether a category is *collected*. Declared in code
  (`CURRENCY_ROUTED_CATEGORIES = [OFFERING]`), not inferred from configuration, so it holds
  before anything is mapped.
- A collected gift carries **no QuickBooks customer at all** — `CustomerRef` is omitted from
  the payload (it is optional on a sales receipt) and the giver's name is kept off the line
  description. It needs no matched contact and no approval.
- It books against a **currency-specific giving item** (`OFFERING:UGX` → QuickBooks Item), which
  is what credits the offering income account. Required; posting blocks until it is mapped.
- An optional **deposit-account override** per category and currency, for churches whose
  offerings land in a different bank account. Only Bank / Other Current Asset accounts are
  offered, and an invalid one is rejected when chosen rather than by Intuit at posting time
  (error 6000). Left unset, offerings deposit where the statement was imported.
- Location and FOB fall back to the tenant's mother group when there is no giver, flagged as a
  fallback on the receipt.

**Categorisation**
- The import wizard now assigns the item mapped to the category it detected, instead of the
  wizard's default item, when the statement message names no item.
- `POST /transactions/import/:accountId` (direct file upload) now sets a category. It never did;
  only the wizard path categorised.
- `PUT /transactions` accepts `externalItemId`, deriving `category` from it. The posting plugin
  prefers the stored item over the category mapping, so correcting only the category would have
  changed nothing in the books.
- `PATCH /transactions/giving-item` re-categorises several rows under one QuickBooks lookup.
- Both fields are frozen once a gift is `POSTED`; correcting a posted gift belongs in QuickBooks.

**Matching**
- Collected categories are excluded from the matching sweep — there is no giver to find.
- A person's name in the statement message is tried before the statement's own sender name,
  falling back to it when the message matches nobody. Givers regularly send on another person's
  phone.

**Permissions**
- New `INTEGRATIONS_VIEW` / `INTEGRATIONS_EDIT`, so connecting QuickBooks no longer requires
  admin rights.
- `PermissionsGuard` applied across all finance controllers and the QuickBooks controller:
  reads require `*_VIEW`, writes `*_EDIT`. The OAuth callback stays `@Public()`.

**No migrations.** Every new mapping is a row in the existing generic `external_system_mapping`
table, keyed by `internalReferenceType = 'CATEGORY_CURRENCY'`.

# How should this be tested?

Automated: `npm test` — 55 suites, 501 tests. `npx tsc --noEmit` clean.

Manually, against a QuickBooks sandbox with at least one giving item:

1. **Offertory end to end.** Import a MoMo statement containing an offering row. It should appear
   as PENDING with "No giver needed", selectable and postable with no match and no approval.
   First post blocks with `CATEGORY_CURRENCY_ITEM_MISSING`; pick the offering item once per
   currency in the dialog, then post. Confirm in QuickBooks that the receipt has **no customer**
   and credits the offering income account.
2. **Tithe unchanged.** A tithe row still requires a matched, approved giver and posts under that
   customer, depositing to the imported account.
3. **Deposit override.** Try mapping an Income account as the deposit — it should be refused in
   the dialog naming the account and its type. Then choose "Deposit where the statement was
   imported (recommended)", which removes the override.
4. **Category correction.** Change a row's category from the table dropdown; confirm the item and
   category move together. Change a row to an offering category and confirm it becomes postable
   without a match. Attempt to change a posted row — it should be refused.
5. **Mixed statement.** Import one file containing tithe, offering, donation and Arise & Build and
   confirm each row carries its own category and item.
6. **Permissions.** With a non-admin user holding no finance permissions, confirm finance
   endpoints return 403; grant `FINANCE_VIEW`/`FINANCE_EDIT` and confirm access returns.

# Notes for reviewers

- **Rollout: permissions must be granted before deploy is complete.** The new guards enforce
  `FINANCE_*` and `INTEGRATIONS_*` on endpoints that previously accepted any authenticated
  tenant user. Admins are unaffected (`roleAdmin` holds every permission), but a finance user
  without the explicit permission will get 403. Grant the permissions to the relevant roles as
  part of the release, and note `INTEGRATIONS_*` is new so no role has it yet.
- **One behaviour change to an existing test expectation.** `momo-import.spec.ts` previously
  asserted that rows naming no QuickBooks item got `externalItemName: null`. They now inherit
  their category's item, which is the point of the wizard fix. The expectation was updated
  deliberately, not to make a failure go away.
- **`PATCH /transactions/giving-item` currently has no client caller.** The bulk category control
  was removed from the reconciliation screen on request, as it did not belong in the
  approve-and-post flow. The endpoint and its tests are retained because the capability is
  wanted; flag it if you would rather it were removed until a UI needs it.
- **Known design smell, not introduced here but worsened by it.** `getSetup` fetches each
  QuickBooks reference list behind its own condition (`!customerMap`, `!itemMap`, `!accountMap`).
  Adding a picker means remembering to widen the right one, and that caused three separate
  "empty dropdown" bugs during development. Driving the fetches off which pickers are actually
  being offered would remove the trap. Left alone here to keep the diff to the feature.
- **`preflight` is query-heavy per transaction** and `postMany` runs it per row. This change adds
  one or two more lookups to that path. Serial posting is deliberate (Intuit rate limits), but
  resolving mappings once per batch is worth doing before large statements are routine.
- The `CategoryRoutingService` rename (from `CategoryCustomerService`) touches ~10 files
  mechanically; it now resolves items and deposit accounts, not just customers.

# Out of scope

- **Per-campus attribution for offertory.** With no giver, Location and FOB fall back to the
  mother group, so offertory does not split by campus in QuickBooks. The statement message does
  carry the campus code (`WHARUAYXPOFFERTORY` → `YXP`), but it is parsed and discarded today;
  using it needs a column and therefore a migration. Raised for a separate decision.
- **Encrypting QuickBooks tokens.** `external_system_connection.accessToken` / `refreshToken` are
  stored in plaintext. A refresh token grants full ledger read/write for 100 days. Pre-existing;
  flagged in the earlier review and still worth scheduling.
- **The in-memory OAuth `state` store**, which does not survive a restart or a second instance.
- **`createUserDto` in `auth.helpers.ts`** discards its `permissions` array (`concat` result
  unused). Masked today because `generateToken` overwrites the field, but it will mislead the
  next caller that trusts that DTO.
- **Extending narration parsing to phone numbers** for matching. Only the name is used; tithe
  number was already supported.

---

## Summary by CodeRabbit

- **New Features**
  - Added currency-based accounting for collected offerings, including category-specific items and deposit accounts.
  - Added bulk transaction recategorization and improved item resolution during imports and updates.
  - Added role-based permissions across finance and QuickBooks operations.
  - Added clearer accounting attribution for collected offerings.

- **Bug Fixes**
  - Improved transaction matching using narration names and excluded currency-routed offerings.
  - QuickBooks token refresh now prompts reconnection for invalid grants.

- **Removals**
  - Removed the QuickBooks card-charging feature.


iamMarybrightEE:
@coderabbitai resume

iamMarybrightEE:
## Automated triage — PR #296

**All six gates pass. Routing: Senior Engineering.**

This touches `src/migrations/**` [schema — auto-runs against staging on merge], `src/finance/**` [real money] and `src/auth/**` [permissions]. Any one of those routes here on its own. The work looks sound; the routing is about what the diff touches, and about two things below that a senior should see before this lands.

**Disclosure:** I wrote the code in this PR. Routing is decided by the diff, not by who ran the skill, so that does not change the outcome — but a self-review is worth declaring, and it is a reason to read the findings rather than the "all gates pass" line.

---

### Gate results

| # | Gate | Result |
|---|------|--------|
| 1 | CodeRabbit reviewed | **PASS** — 4 reviews, all `COMMENTED` (it never approves; that's normal) |
| 2 | All review threads resolved | **PASS** — 10 of 10 resolved, none outdated |
| 3 | CI green | **PASS** — `test` passed (1m4s), CodeRabbit "Review completed". Base is `develop`, so no CodeQL run is expected |
| 4 | Non-trivial description | **PASS** on substance, but it contains one statement that is now false — see below |
| 5 | No credential-shaped files | **PASS** — nothing matching `.env*`, `*.pem`, `*.key`, `id_rsa*`, `*.dump`, `*credentials*`, `*secrets*` |
| 6 | Lockfile | **N/A** — `package-lock.json` untouched, `package.json` unchanged. No dependencies added |

---

### Two things to settle before merge

**1. The description says "No migrations." That is no longer true.**

This PR carries `src/migrations/1789661943339-WidenPersonNameFields.ts`, and `origin/develop` does not have it — verified, not assumed. It arrived through a merge from a develop state that has since moved on, which is exactly how a migration gets into a PR without anyone deciding to put it there.

It matters because merging deploys staging and auto-runs `npm run migration:run`. A reviewer who trusts the description will not be expecting a schema change at all.

The migration itself is benign and worth confirming rather than taking on trust:

```sql
ALTER TABLE "person" ALTER COLUMN "firstName"  TYPE character varying(100)
ALTER TABLE "person" ALTER COLUMN "lastName"   TYPE character varying(100)
ALTER TABLE "person" ALTER COLUMN "middleName" TYPE character varying(100)
```

Widening a `varchar` in place. It rewrites no data, cannot fail on existing rows, and takes only a brief lock. Safe to deploy.

Its `down()` narrows back to `varchar(40)`, which **will fail** if any name longer than 40 characters has been stored by then — Postgres refuses the shrink rather than truncating. That is the right failure (refusing beats silently cutting someone's name), but it means the rollback is not guaranteed to work once the wider column is in use. Worth knowing before anyone reaches for `migration:revert`.

Please correct the "No migrations" line so the next reader is not misled.

**2. 62 routes across 10 controllers start requiring a permission on merge. (Rollout, not code.)**

`PermissionsGuard` is applied across every finance controller and the QuickBooks controller — reads need `*_VIEW`, writes `*_EDIT`. These endpoints previously accepted any authenticated tenant user.

Administrators are unaffected: `roleAdmin` holds every permission in `permissionsList`. Everyone else gets a 403 until the permission is granted to their role. **`INTEGRATIONS_VIEW` and `INTEGRATIONS_EDIT` are created by this PR, so no role holds them yet** — Integrations disappears for every non-admin the moment this lands.

This is disclosed in the PR body and it is the correct direction of travel: the handoff doc already tracks "enforce FINANCE_VIEW/FINANCE_EDIT on finance controllers" as outstanding. It just needs the role grants to happen in the same window as the deploy, not after it.

---

### Review findings

**Multi-tenant correctness: clean.** `CategoryRoutingService` is new and reaches no repository directly — every read goes through `ExternalSystemMappingService`, which calls `requireTenant()` in all seven of its methods. No new query can return another tenant's rows.

**The OAuth callback stays reachable.** `@Public()` routes are deliberately skipped when the permission decorators are applied, so `GET /api/integrations/quickbooks/callback` — which Intuit hits as a browser redirect with no JWT — still works. That is the one route where a guard would have broken the whole connect flow, and it is worth checking by eye rather than trusting.

**`PATCH /api/finance/transactions/giving-item` has no caller.** The bulk category control was removed from the reconciliation screen on request, since it did not belong in the approve-and-post flow. The endpoint and its tests remain because the capability is wanted. Flag it if you would rather it came out until a UI needs it — it is defensible either way, but it should be a decision rather than an oversight.

**`POST /api/integrations/quickbooks/charges` gained a required field.** `CreateChargeDto.requestId` is now mandatory; a request without it is a 400. Nothing in the client calls this endpoint — `quickbooksCharges` is defined in `remoteRoutes` but never used — so nothing in the app breaks. Anything external (a Postman collection, a script) needs the new field. The reason it is required rather than generated: Intuit treats `Request-Id` as the idempotency key for a charge, so a retry with a fresh value charges the card a second time.

**`preflight` makes a QuickBooks call per transaction when a deposit override is mapped. (CodeRabbit did not flag this.)**

`preflight` calls `getQboAccounts` to validate a mapped deposit account, and `postMany` runs `preflight` once per row. A batch of 300 offerings with a deposit override mapped therefore makes 300 identical calls to Intuit, all fetching the same chart of accounts.

It matters because Intuit rate-limits per realm, and this sits on the money path — the failure mode is a throttled batch part-way through, which is exactly when you least want one. It is bounded today: the call only fires when a deposit override exists, and offerings are expected to leave that unset, so most batches make zero extra calls. But it is a sharp edge that only shows up at volume.

The fix is to resolve the account list once per batch rather than once per row, or to drop the preflight check now that `applySetup` refuses an invalid account at the moment it is chosen. Not a merge blocker; worth a follow-up before large statements are routine.

**Minor, non-blocking.**

- The description's test counts are stale — it says 55 suites / 501 tests; the branch is now 56 / 516. Harmless, but the numbers are the sort of thing readers use to decide whether to re-run.
- `getSetup` fetches each QuickBooks reference list behind its own condition (`!customerMap`, `!itemMap`, `!accountMap`, `collected`). Adding a picker means remembering to widen the right one, and that pattern caused three separate "empty dropdown" bugs during development. Driving the fetches off which pickers are actually being offered would remove the trap. Left alone here to keep the diff to the feature.

---

### What I verified

- `npx tsc --noEmit` clean; 56 suites / 516 tests passing locally, matching the CI run.
- No credential-shaped files, no dependency change, no lockfile movement.
- Every new mapping is a row in the existing generic `external_system_mapping` table keyed by `internalReferenceType = 'CATEGORY_CURRENCY'` — no schema was added for the feature itself. The one migration present is the unrelated name-widening described above.
- The `person` name widening is consistent between entity and migration: the entity declares `length: 100` and the migration alters to `varchar(100)`, so schema sync has nothing left to reconcile after it runs.

---

### To merge

Senior Engineering review, then approve and merge in the GitHub UI — this skill never does either. Before merging: fix the "No migrations" line, and line up the role grants (including the brand-new `INTEGRATIONS_*` pair) so the permission enforcement does not lock finance staff out on deploy.

iamMarybrightEE:
@kakoma needs senior engineering review

kakoma:
## PR Review — Finance Module: Offertory Posting, Permissions & Category Correction

**Routing reason:** This PR touches `src/auth/constants.ts` (new permission constants) and multiple `src/finance/**` files (real money) — both are sensitive paths. Routed to the Senior Engineering Team for review. This does not mean the PR is problematic; it means these paths always get a senior pass.

---

### Gate checks

| Gate | Result |
|------|--------|
| CodeRabbit reviewed | ✅ Pass — 4 review passes by `coderabbitai` |
| All review threads resolved | ✅ Pass — all 8 threads resolved (including 3 that are also outdated) |
| CI green | ✅ Pass — `test` passing, CodeRabbit "Review completed" |
| Non-trivial description | ✅ Pass — detailed description with test plan, design notes, out-of-scope items |
| No credential-shaped files | ✅ Pass |
| Lockfile | N/A — no `package-lock.json` or `package.json` changes |

---

### Adversarial review

**Verdict:** This is a substantial, well-constructed PR. The core offertory routing logic is sound — collected categories correctly omit the customer from the QBO sales receipt, the currency-specific item lookup blocks before posting rather than at Intuit's error 6000 fault, and the deposit account override is optional and validated at setup time. The permission guard rollout across all finance and QBO controllers is systematic and correct. The most significant issues (deposit readiness blocking `ready`, rule-match item application, OAuth connect requiring EDIT permission, request-ID reuse across retries) were caught and fixed during the review cycle. The remaining findings below do not prevent merge but should be reviewed by a senior engineer.

---

### Findings

**1. `bulkUpdateGivingItem` silently clears all categories when `externalItemId` is omitted**

`BulkUpdateGivingItemDto.externalItemId` is `@IsOptional()`. A request body of `{ "transactionIds": [1, 2, 3] }` with no `externalItemId` is valid. The service then calls `resolveGivingItem(undefined ?? null)` = `resolveGivingItem(null)`, which returns `{ externalItemId: null, externalItemName: null, category: DEFAULT_TRANSACTION_CATEGORY }`. All three transactions are silently reset to the default category with no giving item.

This is a data integrity trap. The PR description notes this endpoint currently has no client caller, which makes it more dangerous: the first developer who calls it without `externalItemId` (expecting a no-op or a 400) will corrupt categories in bulk.

**Fix:** Either make `externalItemId` required in `BulkUpdateGivingItemDto` (removing `@IsOptional()`), or add a guard that returns early or throws a 400 when `externalItemId` is absent — "bulk-clear categories" is almost certainly never the intent of this endpoint.

---

**2. `requiresMatch` injected onto `Transaction` entity without a DTO — undocumented API contract change**

In `transactions.service.ts`, both `update()` and `findAll()` return `{ ...saved, requiresMatch: ... } as Transaction`. The `Transaction` entity has no `requiresMatch` field; the `as Transaction` cast silences TypeScript. The client will start receiving this field on every transaction response.

CodeRabbit did not flag this. It matters because if the client TypeScript types mirror the server entity, the client will not see this new field without an update to client types. If the client performs exhaustive type checks it may silently discard it.

**Fix:** This is minor if a companion client PR is already updating the types. If not, add a note tracking the client-side type update, or define a `TransactionResponseDto` that includes `requiresMatch?: boolean` and return that type from the affected endpoints.

---

**3. Existing QBO connections need manual reconnection before the Payments endpoint works — rollout notes incomplete**

The `POST /integrations/quickbooks/charges` endpoint requires `com.intuit.quickbooks.payment` scope. Any existing connection was authorized with only `com.intuit.quickbooks.accounting`. The first `createCharge` call will get a 403 from Intuit, which `isPaymentsPermissionFailure` correctly catches and converts to `ForbiddenException("Please disconnect QuickBooks and connect again to enable payment processing.")`. The failure is handled gracefully.

The rollout notes in the PR description do not mention this. An operator enabling this feature on a live tenant will find the payment endpoint failing with an unclear message until they realize they need to disconnect and reconnect QBO.

**Fix:** Add to the rollout notes: *"Existing QBO connections must be disconnected and reconnected (QBO Settings → Disconnect → Connect again) before `POST /charges` will work, because the OAuth grant must be extended to include `com.intuit.quickbooks.payment`."*

---

**4. Deployment will 403 every non-admin finance user until roles are updated — no automated mitigation**

Every finance controller and the QBO controller now require `FINANCE_VIEW/EDIT` or `INTEGRATIONS_VIEW/EDIT`. Previously, any authenticated tenant user could access them. No migration seeds these permissions to existing roles. This is a breaking deploy for any tenant that has non-admin finance users.

The PR description explicitly flags this and notes admins are unaffected. The risk is that it requires coordinated manual action at deploy time with no automated fallback.

CodeRabbit did not flag this because it is a deployment concern rather than a code bug. Noting it here so Senior Engineering can confirm the deploy checklist includes the permission grant step before this hits staging or production.

---

### What CodeRabbit missed

Finding 1 (silent bulk category clear) was not raised. Findings 2–4 are operational and would not typically appear in static analysis.

---

### Priority ranking

1. **Finding 1 is the only code-level concern worth addressing before merge** — silent data corruption on a bulk-write endpoint with no current UI caller is exactly the kind of issue discovered when a developer writes an integration test or a CLI script and passes a partial body. Since the endpoint has no client yet, fixing the DTO now costs nothing.
2. Findings 2–4 are operational/communication items. Finding 2 may already be addressed by a companion client PR. Findings 3–4 are deploy notes.

No other blockers. Multi-tenant correctness, the offertory routing model, the permission architecture, and the QBO payment retry-with-idempotency are all sound.

kakoma:
**Merge dependency — frontend must ship first**

The companion frontend PR has been reviewed against this PR's contract. Five of seven questions pass. One is a blocker:

`postsAs` (`'GIVER' | 'CATEGORY'`) is not yet in the frontend preview response type, so the posting dialog renders a "Customer" row for offertory transactions regardless of routing. An operator would see a blank or null customer on a receipt that intentionally carries no customer — misleading at the point of posting.

**This PR should not merge to develop until the frontend fix lands.** The frontend change is small (add `postsAs` to the preview type, suppress the Customer row when `postsAs === 'CATEGORY'`), but it must go first — or in the same deploy window with the frontend PR merged immediately before this one.

